PR #5, opened 2016-06-20
alias pg to pg:info and add argument to filter by db

jdx:


jdx:
since this brings in the addons resolver we'll reuse a ton in this plugin, it would be good if you could test that to see if it works for all expected input types we discussed here: https://github.com/heroku/heroku-cli-addons/pull/25 (it should)

commands/info.js
```diff
@@ -1,14 +1,14 @@
...
-let co = require('co')
-let cli = require('heroku-cli-util')
-let _ = require('lodash')
-let url = require('url')
-let host = require('../lib/host')
+const co = require('co')
+const cli = require('heroku-cli-util')
 
 function databaseNameFromUrl (uri, config) {
+  const invert = require('lodash.invert')
```
msakrejda: Out of curiosity, is this some require magic, or does lodash actually ship micro-modules so you can avoid requiring the whole thing?

jdx: No magic, they just ship the micro modules. Lodash as a whole takes ~40ms to load I've found, so I avoid loading the whole thing.

jdx: Though given how slow `/addons` is it's not like it makes a big difference, I'm just doing this out of convention.

commands/info.js
```diff
@@ -28,17 +28,24 @@ function displayDB (db) {
...
   let config = heroku.get(`/apps/${app}/config-vars`)
-  let addons = yield heroku.request({
-    method: 'get',
-    path: `/apps/${app}/addons`
-  })
-  addons = addons.filter((a) => a.addon_service.name === 'heroku-postgresql')
-  addons = _.sortBy(addons, notDatabaseUrl, 'name')
+
+  if (db) {
+    addons = yield [fetcher.addon(heroku, app, db)]
```
msakrejda: If the addon is not found, does the fetcher error out and exit here?

commands/info.js
```diff
@@ -52,18 +59,17 @@ function * run (context, heroku) {
...
   topic: 'pg',
   needsApp: true,
   needsAuth: true,
+  args: [{name: 'database', optional: true}],
   run: cli.command({preauth: true}, co.wrap(run))
 }
 
 exports.displayDB = displayDB
+exports.root = cmd
+exports.info = Object.assign({}, cmd, {command: 'info'})
```
msakrejda, line 89: Nice.

commands/info.js
msakrejda: Oh I see, fetcher is new below. So it doesn't. What happens when the addon is not found?

jdx: it shows an error from the API

jdx:
```
$ heroku pg:info wlekjflwekjflkwejf -a forker
 ▸    Couldn't find that add-on.
```

lib/fetcher.js
```diff
@@ -0,0 +1,30 @@
...
+
+function * addon (heroku, app, db) {
+  const {resolve} = require('heroku-cli-addons')
+  let attachment = yield resolve.attachment(heroku, app, db)
+  return yield heroku.get(`/addons/${attachment.addon.id}`)
+}
+
+function * all (heroku, app) {
+  const sortBy = require('lodash.sortby')
+  const notDatabaseUrl = (a) => !a.config_vars.find((c) => c === 'DATABASE_URL')
```
msakrejda: Maybe out of scope here, but should we have a secondary sort order by addon name or attachment name or something?

jdx: it is doing that, by passing in `'name'` to `sortBy`

lib/fetcher.js
```diff
@@ -0,0 +1,30 @@
...
+
+  // TODO: make this more efficient without having to call /addons
+  // (will require an API change of some sort)
+  let {attachments, addons} = yield {
+    attachments: heroku.get(`/apps/${app}/addon-attachments`),
+    addons: heroku.get('/addons')
+  }
+  addons = addons.filter(a => a.addon_service.name === 'heroku-postgresql')
```
msakrejda: Can we support staging add-ons here? This would filter them out. Right now support is kinda janky based on env vars (I think `HEROKU_POSTGRESQL_HOST` and `HEROKU_POSTGRESQL_ADDON_NAME` or something like that), but it works. I think something more robust would be harder, but porting the existing mechanism would be great.

lib/fetcher.js
msakrejda: Oh wait that's exactly what you are doing below.

lib/fetcher.js
jdx: can you point me to that logic? having trouble finding it

msakrejda: Sure. There's basically three things that are different for staging add-ons:
- the addon slug (`heroku-postgresql-maciek`)
- the addon prefix (`HEROKU_POSTGRESQL_MACIEK` as in `*HEROKU_POSTGRESQL_MACIEK*_PUCE_URL`)
- the api endpoint (`shogun-maciek.herokuapp.com`)

I _think_ we've eliminated the dependency on the second one in the CLI, but the first is handled [here](https://github.com/heroku/heroku/blob/master/lib/heroku/helpers/heroku_postgresql.rb#L120) and the third [here](https://github.com/heroku/heroku/blob/master/lib/heroku/client/heroku_postgresql.rb#L26) and [here](https://github.com/heroku/heroku/blob/master/lib/heroku/client/heroku_postgresql_backups.rb#L35) (the method those two call is below). I think you can drop support for the alternate `SHOGUN` env var--that's an obsolete legacy mechanism.

jdx: on that note, can any of this logic be removed? https://github.com/heroku/heroku-pg/blob/master/lib/host.js

jdx: when you say addon slug you mean `a.addon_service.name` right?

jdx: which means all I would need to do to fix this would be to check for the value of `process.env.HEROKU_POSTGRESQL_ADDON_NAME` if it exists instead of `heroku-postgresql`

msakrejda: Yeah. It may be worth centralizing that somewhere as you build out other commands but maybe not.

msakrejda: Oh I see and you're already handling host--yeah, drop the `SHOGUN` stuff from that file and unfortunately the rest has to stay for now.

msakrejda:
Something is not right here:

``` console
maciek@mothra:~/code/heroku-pg$ h pg:info --app shogun
Two-factor code: ********************************************
=== glowing-gladly-2371
Config Vars:        DATABASE_URL, HEROKU_POSTGRESQL_GOLD_URL
Plan:               Standard 2
Status:             Available
Data Size:          11.05 GB
Tables:             2
PG Version:         9.4.1
Connections:        2/400
Fork/Follow:        Available
Rollback:           earliest from 2016-06-17 20:03 UTC
Created:            2015-05-01 02:53 UTC
Region:             virginia
Maintenance:        not required
Maintenance window: Fridays 18:30 to 22:30 UTC

=== postgresql-opaque-35949
Config Vars:        DATABASE_URL, HEROKU_POSTGRESQL_GOLD_URL
Plan:               Standard 7
Status:             Available
Data Size:          214.64 GB
Tables:             70
PG Version:         9.5.0
Connections:        302/500
Fork/Follow:        Available
Rollback:           earliest from 2016-06-17 20:03 UTC
Created:            2016-02-02 18:50 UTC
Region:             virginia
Followers:          HEROKU_POSTGRESQL_BRONZE
Maintenance:        not required
Maintenance window: Fridays 20:00 to Saturdays 00:00 UTC

=== postgresql-cubed-50807
Config Vars:        HEROKU_POSTGRESQL_BRONZE_URL
Plan:               Standard 7
Status:             Available
Data Size:          214.64 GB
Tables:             70
PG Version:         9.5.2
Connections:        2/500
Fork/Follow:        Unavailable on followers
Rollback:           earliest from 2016-06-17 20:02 UTC
Created:            2016-04-27 17:33 UTC
Region:             virginia
Following:          HEROKU_POSTGRESQL_GOLD
Behind By:          17 commits
Maintenance:        not required
Maintenance window: Fridays 20:30 to Saturdays 00:30 UTC
```

jdx:
alright @uhoh-itsmaciek, I think this is good now.

commands/info.js
```diff
@@ -0,0 +1,89 @@
...
+    if (config[key] === value) keys.push(key)
+  }
+  return sortBy(keys, k => k !== 'DATABASE_URL', 'name')
+}
+
+function databaseNameFromUrl (uri, config) {
+  const invert = require('lodash.invert')
+  const url = require('url')
+
+  delete config.DATABASE_URL
```
msakrejda: I know it doesn't matter to your only caller, but it seems like a future footgun to mutate your argument here. Can you do something like `let conf = Object.assign({}, config)` and then use `conf`?

jdx: yeah that's a much better idea

lib/host.js
```diff
@@ -1,13 +1,12 @@
 'use strict'
 
-module.exports = function (attachment) {
-  let shogun = process.env.SHOGUN
+module.exports = function (addon) {
   let appHost = process.env.HEROKU_POSTGRESQL_HOST
-  if (attachment.plan.name.match(/dev|basic/)) {
+
+  if (addon.plan.name.match(/dev|basic/)) {
```
msakrejda: For what it's worth, this could be `.endsWith('dev') || .endsWith('basic')` to minimize the chance it inadvertently catches some new unrelated plan in the future, but it's probably fine as is.

package.json
```diff
@@ -34,6 +40,6 @@
   "scripts": {
     "postversion": "npm publish && git push && git push --tags",
     "preversion": "npm test",
-    "test": "nyc mocha && standard"
+    "test": "TZ=utc nyc mocha && standard"
```
msakrejda, line 42: Nice.

commands/info.js
jdx: actually I was able to reuse the logic from above here

msakrejda:
+1

lib/host.js
jdx: I'll change it to be more explicit anyhow

PR #9, opened 2016-08-02
added psql function

jdx:
Also exported fetcher so it can be used with `heroku-pg-extras`.

### DIFF
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -31,7 +31,7 @@ function * exec (db, query) {
     if (err.code !== 127) throw err
     cli.error(`The local psql command could not be located.
 For help installing psql, see https://devcenter.heroku.com/articles/heroku-postgresql#local-setup`)
-    process.exit(127)
+    process.exit(1)
   }
 }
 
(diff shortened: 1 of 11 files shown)


dickeyxxx:
needs a test for `lib/fetcher.js`

jdx:
@uhoh-itsmaciek can you take a look at this one? We need it for pg-extras

lib/psql.js
```diff
@@ -0,0 +1,40 @@
...
+const debug = require('debug')('psql')
+
+function * exec (db, query) {
+  const execa = require('execa')
+  const os = require('os')
+  const cli = require('heroku-cli-util')
+
+  debug(query)
+  let env = {
+    PGAPPNAME: 'psql non-interactive',
```
msakrejda, line 13: So stupid question but is this only for the non-interactive commands? How does `pg:psql` work now?

jdx, line 13: yeah this is only partial functionality, and it isn't actually adding anything to this plugin. It's just used by heroku-pg-extras that imports this. I decided to put it here since I know I'll reuse the logic soon for the heroku-pg commands that need to run remote queries.

I still need to write the interactive version for `pg:psql`.

msakrejda, line 13: Ok, cool, got it.

msakrejda:
+1

PR #13, opened 2016-08-10
use execa.shell to run psql to fix windows problems

jdx:


msakrejda:
For what it's worth, this works for me on Linux.

keiko713:
can we ship this? seems like it'll fix this:
https://github.com/heroku/heroku-pg-extras/issues/132

PR #19, opened 2016-08-30
fix psql on windows by not using execa

jdx:


msakrejda:
Looks good, and works for me in Linux.

PR #26, opened 2016-10-08
added pg:backups:url

jdx:


commands/backups/download.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  let info = yield heroku.post(`/client/v11/apps/${app}/transfers/${num}/actions/public-url`, {host})
+  yield download(info.url, output, {progress: true})
+}
+
+module.exports = {
+  topic: 'pg',
+  command: 'backups:download',
+  description: 'downloads database backup',
```
msakrejda, line 49: Nice.

commands/backups/url.js
```diff
@@ -0,0 +1,42 @@
...
+  needsApp: true,
+  needsAuth: true,
+  args: [{name: 'backup_id', optional: true}],
+  run: cli.command({preauth: true}, co.wrap(run))
+}
+
+module.exports = [
+  Object.assign({command: 'backups:url'}, cmd),
+  Object.assign({command: 'backups:public-url'}, cmd),
+  Object.assign({command: 'backups:publicurl'}, cmd)
```
msakrejda: Do all these show up in topic help?

commands/backups/url.js
```diff
@@ -0,0 +1,42 @@
...
+    if (!num) throw new Error(`Invalid Backup: ${args.backup_id}`)
+  } else {
+    let transfers = yield heroku.get(`/client/v11/apps/${app}/transfers`, {host})
+    let lastBackup = sortBy(transfers.filter(t => t.succeeded && t.to_type === 'gof3r'), 'created_at').pop()
+    if (!lastBackup) throw new Error(`No backups on ${cli.color.app(app)}. Capture one with ${cli.color.cmd('heroku pg:backups:capture')}`)
+    num = lastBackup.num
+  }
+
+  let info = yield heroku.post(`/client/v11/apps/${app}/transfers/${num}/actions/public-url`, {host})
+  cli.log(info.url)
```
msakrejda, line 26: We're no longer showing the expiration? /cc @neovintage

lib/download.js
```diff
@@ -0,0 +1,36 @@
...
+    let total = 0
+    rsp.on('data', function (chunk) {
+      total += chunk.length
+      bar.tick(chunk.length, {data: bytes(total, {decimalPlaces: 2, fixedDecimals: 2})})
+    })
+  }
+
+  return new Promise(function (resolve, reject) {
+    mkdirp.sync(require('path').dirname(path))
+    let file = fs.createWriteStream(path)
```
msakrejda, line 26: This silently overwrites the target file--is that a good idea?

lib/download.js
```diff
@@ -0,0 +1,36 @@
...
+    https.get(url, function (rsp) {
+      if (tty && opts.progress) showProgress(rsp)
+      rsp.pipe(file)
```
msakrejda, line 29: Neat.

test/commands/backups/download.js
```diff
@@ -0,0 +1,45 @@
...
+    pg.done()
+  })
+
+  context('with no id', () => {
+    beforeEach(() => {
+      pg.get('/client/v11/apps/myapp/transfers').reply(200, [
+        {succeeded: true, to_type: 'gof3r', num: 3}
+      ])
+    })
+    it('shows URL', () => {
```
msakrejda: Bad test description.

test/commands/backups/download.js
```diff
@@ -0,0 +1,45 @@
...
+      return cmd.run({app: 'myapp', args: {}, flags: {output: './tmp/latest.dump'}})
+      .then(() => expect(fs.readFileSync('./tmp/latest.dump', 'utf8'), 'to equal', '{}'))
+    })
+  })
+
+  context('with id', () => {
+    it('shows URL', () => {
```
msakrejda: Ditto.

commands/backups/url.js
jdx: @uhoh-itsmaciek so this was a design decision on my part. Before `pg:backups:url` would act different depending on whether or not it was a tty but I feel it's better to be consistent. Thoughts? I could add expiration under a flag or something possibly.

commands/backups/url.js
jdx: yeah that wasn't intended though, which do you think is best? I like `heroku pg:backups:url` personally

lib/download.js
jdx: think it would be good to add a mechanism to add a suffix to the file? like `latest.dump.1`?

I imagine if the user specifies the file it should overwrite though.

commands/backups/url.js
msakrejda: I'm fine with the change if Rimas is. "Act differently depending on whether output is tty" is not unprecedented (e.g., GNU `ls` does this for `--color=auto` which is now default--not sure about BSD `ls`), but I don't feel strongly one way or the other. I just don't want to sneak in product changes. I'll ping him again.

lib/download.js
msakrejda: I suppose. I think the easiest would be to just error if the file exists. I think downloading backups is uncommon enough that it's not likely to become annoying.

commands/backups/url.js
jdx: agreed, and it's not without precedent (we already do this with color all the time), but we'll see what he says. In terms of impact I'm not that worried about this since the automated case is staying the same so it won't break any scripts or anything.

commands/backups/url.js
msakrejda: Agreed, `pg:backups:url` should be the main one--I see you changed the other ones to be hidden already.

jdx: took a gamble 🎲

lib/download.js
jdx: I went ahead and put the logic in for the suffixes

msakrejda (approved):

PR #28, opened 2016-10-08
added pg:backups:restore

jdx:


commands/backups/restore.js
```diff
@@ -0,0 +1,88 @@
+'use strict'
+
+const co = require('co')
+const cli = require('heroku-cli-util')
+
+function dropboxURL (url) {
+  // Force a dump file to download instead of rendering as HTML
+  // file by specifying the dl=1 param for Dropbox URL
```
msakrejda, line 8: Wat? TIL. Nice work @keiko713.

commands/backups/restore.js
```diff
@@ -0,0 +1,88 @@
...
+  if (url.match(/www.dropbox.com/) && !url.endsWith('dl=1')) {
```
msakrejda: Should we do more principled filtering here, though? This will match things like `https://example.com?q=www.dropbox.com`. I suppose it's not likely to actually have false-positive matches in the real world, though, so maybe we can leave it to future cleanup.

commands/backups/restore.js
```diff
@@ -0,0 +1,88 @@
...
+    if (url.endsWith('dl=0')) url = url.replace('dl=0', 'dl=1')
+    else if (url.includes('?')) url += '&dl=1'
```
msakrejda, line 11: Note that an unencoded `?` can appear in the password or username field (e.g., `https://?:?@dropbox.com/foo`). Again, probably not a real-world problem.

msakrejda (approved):

commands/backups/restore.js
keiko713: haha yeah, the customer had this issue and debugging what was causing this was fun. but as Maciek commented below, the solution isn't super good, so any cleanup would be nice, but for now maybe we can just leave it.

commands/backups/restore.js
jdx: I fixed this one, which I think pretty much makes the second issue unneeded since you wouldn't be using http auth with dropbox anyways

msakrejda: Works for me.

PR #31, opened 2016-10-09
added pg:backups:schedule

jdx:
needs a test

### DIFF
--- a/commands/backups/schedule.js
+++ b/commands/backups/schedule.js
@@ -1,0 +1,65 @@
+'use strict'
+
+const co = require('co')
+const cli = require('heroku-cli-util')
+
+const TZ = {
+  'PST': 'America/Los_Angeles',
+  'PDT': 'America/Los_Angeles',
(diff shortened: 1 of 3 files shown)


commands/backups/schedule.js
```diff
@@ -0,0 +1,64 @@
...
+
+function * run (context, heroku) {
+  const host = require('../../lib/host')
+  const fetcher = require('../../lib/fetcher')(heroku)
+
+  const {app, args, flags} = context
+
+  let db = yield fetcher.addon(app, args.database)
+
+  let schedule = parse(flags.at)
```
msakrejda, line 36: Might as well parse this before fetching the DB above--there's no point in continuing if the time is invalid.

msakrejda (changes requested):

commands/backups/schedule.js
jdx: fixed

msakrejda (approved):

PR #33, opened 2016-10-10
added app name to pg info

jdx:


test/commands/info.js
```diff
@@ -82,6 +82,7 @@ describe('pg', () => {
       .then(() => expect(cli.stdout, 'to equal', `=== DATABASE_URL, HEROKU_POSTGRESQL_COBALT_URL
 Plan:      Hobby-dev
 Following: HEROKU_POSTGRESQL_COBALT
+App:       myapp2
```
msakrejda: App as in billing app? Should we make that more explicit here? I'm on the fence, but it's not super clear.

msakrejda (approved):

test/commands/info.js
jdx: any thoughts @ransombriggs?

ransombriggs: @uhoh-itsmaciek @dickeyxxx I am down with adding it as `Billing App`

PR #35, opened 2016-10-11
added pg:psql

jdx:


commands/psql.js
```diff
@@ -0,0 +1,33 @@
...
+  wantsApp: true,
+  needsAuth: true,
+  flags: [{name: 'command', char: 'c', description: 'SQL command to run', hasValue: true}],
+  args: [{name: 'database', optional: true}],
+  run: cli.command({preauth: true}, co.wrap(run))
+}
+
+module.exports = [
+  Object.assign({topic: 'pg', command: 'psql'}, cmd),
+  Object.assign({topic: 'psql'}, cmd)
```
msakrejda, line 32: Ooh so we'll have `heroku psql`? Fancy!

lib/psql.js
```diff
@@ -18,17 +22,31 @@ function * exec (db, query) {
...
   return stripEOF(stdout)
 }
 
+function * interactive (db) {
+  const {spawnSync} = require('child_process')
+  let name = db.addon.config_vars[0].replace(/^HEROKU_POSTGRESQL_/, '').replace(/_URL$/, '')
+  let prompt = `${db.addon.app.name}::${name}%R%# `
+  let {error: err, status} = spawnSync('psql',
+    ['--set', `PROMPT1=${prompt}`, '--set', `PROMPT2=${prompt}`],
+    {env: env(db), stdio: 'inherit'})
```
msakrejda, line 49: This means that the `psql` process inherits stdin, so something like `<<<'select 1' heroku psql` should still work, right?

msakrejda: Oh also, do you know the behavior with signals? I think that affects both the interactive and non-interactive cases: e.g., if I have `SELECT pg_sleep(1000000)`, I should be able to kill it in both cases with `Ctrl-C`. I checked the Ruby `pg:psql` and this does work.

commands/psql.js
jdx: yep!

lib/psql.js
jdx: signals are weird, I'll fix that

jdx: alright, I added a patch for SIGINT at least. Do you think I should look into other signals as well?

msakrejda: No idea. Like, it'd be nice for it to do the right thing, but I have no idea how this works in Ruby or how it works in node, and only a vague idea of how it works in principle. So maybe this is good enough for now?

msakrejda (approved):

lib/psql.js
jdx: yeah I think so. The bit that has go -> node also has to propagate the signals and I think right now the only one it does is SIGINT and it doesn't seem to have caused any issues for anyone. Now that I think of it, if we wanted to add more we would have to edit that part of the codebase too.

commands/psql.js
```diff
@@ -0,0 +1,33 @@
...
+const cli = require('heroku-cli-util')
+
+function * run (context, heroku) {
+  const fetcher = require('../lib/fetcher')(heroku)
+  const psql = require('../lib/psql')
+
+  const {app, args, flags} = context
+
+  let db = yield fetcher.database(app, args.database)
+  cli.console.error(`--> Connecting to ${cli.color.addon(db.addon.name)}`)
```
ransombriggs, line 13: @dickeyxxx can I change this to the config var, it used to be this way, and in the case of implicit attachment, there is no addon to speak of.  The use case we are trying to support is someone `config:set` the DATABASE_URL and running `heroku pg:psql` and expecting it to connect.

ransombriggs: @dickeyxxx After some more digging I found that we can always assume an addon.  The old ruby code does not support DATABASE_URL when there is no matching attachment.  It basically pulls the config vars and tries to match up DATABASE_URL with the attachment config vars and if it matches, then it uses that attachment.

jdx: we're trying to move towards using addon names instead of config vars right?

PR #36, opened 2016-10-11
added pg:credentials

jdx:


commands/credentials.js
```diff
@@ -0,0 +1,44 @@
...
+Connection URL:
+   ${db.url.href}`)
+  })
+
+  let reset = co.wrap(function * () {
+    const host = require('../lib/host')
+    let db = yield fetcher.addon(app, args.database)
+    yield cli.action(`Resetting credentials on ${cli.color.addon(db.name)}`, co(function * () {
+      yield heroku.post(`/client/v11/databases/${db.name}/credentials_rotation`, {host: host(db)})
+      // TODO: is the promotion stuff necessary here?
```
msakrejda: I don't think so, but what promotion stuff? If you mean ensuring the app gets the new config vars, no, Shogun (or Yobuko) will post them to the app and the app will restart. Because `DATABASE_URL` is now a real attachment, that will pick up the new value as well.

If dealing with a legacy app where `DATABASE_URL` was just copy-and-pasted, I _think_ API has (had?) some magic to automatically update config vars matching the same value, but I'm not sure about this. There also seems to be an implicit attachment behavior with `config:set` matching a config var right now, so I can't think of a way to test this directly.

test/commands/credentials.js
```diff
@@ -0,0 +1,62 @@
...
+const expect = require('unexpected')
+const nock = require('nock')
+const proxyquire = require('proxyquire')
+
+const db = {
+  database: 'mydb',
+  host: 'foo.com',
+  user: 'jeff',
+  password: 'pass',
+  url: {href: 'postgres://foo'}
```
msakrejda: Not a big deal, but I think internal consistency in the test data would be nice, so `{href: 'postgres://jeff:pass@foo.com/mydb'}`...

msakrejda (approved):

commands/credentials.js
jdx: sounds good to me, removed the TODO

test/commands/credentials.js
jdx: 👍  updated

msakrejda:
+1

commands/credentials.js
```diff
@@ -0,0 +1,43 @@
...
+    yield reset()
+  } else {
+    yield showCredentials()
+  }
+}
+
+module.exports = {
+  topic: 'pg',
+  command: 'credentials',
+  description: 'display the database credentials',
```
msakrejda: As long as we're here, can you `s/display/manage/`? Right now it's not obvious from the topic-level help that this can reset credentials.

jdx: done

msakrejda: :tada:

PR #40, opened 2016-10-11
added pg:upgrade

jdx:


lib/host.js
```diff
@@ -1,9 +1,11 @@
 'use strict'
 
+const util = require('./util')
+
 module.exports = function (addon) {
   let host = process.env.HEROKU_POSTGRESQL_HOST
 
-  if (addon && (addon.plan.name.endsWith('dev') || addon.plan.name.endsWith('basic'))) {
+  if (addon && util.starterPlan(addon)) {
     if (host) return `https://${host}`
```
msakrejda: Wouldn't this be more straighforward as

``` javascript
if (host) {
  return `https://${host}`
} else if (addon && util.starterPlan(addon)) {
  return 'https://postgres-starter-api.heroku.com'
} else {
  return 'https://postgres-api.heroku.com'
}
```

?

msakrejda (approved):

lib/host.js
jdx: 👍  updated

msakrejda:
+1

PR #41, opened 2016-10-11
added pg:links

jdx:


commands/links/destroy.js
```diff
@@ -0,0 +1,34 @@
...
+    yield heroku.delete(`/client/v11/databases/${db.name}/links/${encodeURIComponent(args.link)}`, {host: host(db)})
+  }))
+}
+
+module.exports = {
+  topic: 'pg',
+  command: 'links:destroy',
+  description: 'destroys a link between data stores',
+  help: `Example:
+  heroku pg:links:create HEROKU_POSTGRESQL_CERULEAN redis_symmetrical_100`,
```
msakrejda: this should be `pg:links:destroy`, no? Also the link target name is not a standard format we use--should this be something like `redis-symmetrical-100` instead?

test/commands/links/create.js
```diff
@@ -0,0 +1,45 @@
...
+    nock.cleanAll()
+    api.done()
+    pg.done()
+  })
+
+  it('creates a link', () => {
+    pg.post('/client/v11/databases/postgres-1/links', {target: 'postgres-1', as: 'foobar'}).reply(200, {name: 'foobar'})
+    return cmd.run({app: 'myapp', args: {}, flags: {confirm: 'myapp', as: 'foobar'}})
+    .then(() => expect(cli.stdout, 'to equal', ''))
+    .then(() => expect(cli.stderr, 'to equal', 'Adding link from postgres-1 to postgres-1... foobar\n'))
```
msakrejda: This reads kinda weird... should we  change this to `... done; added foobar` or `... added foobar` or something? /cc @neovintage

msakrejda (changes requested):

commands/links/destroy.js
jdx: yes to both, fixed

test/commands/links/create.js
jdx: 👍  to making it more clear, updated

msakrejda (approved):

PR #43, opened 2016-10-11
added pg:push and pg:pull

jdx:


commands/pull.js
```diff
@@ -0,0 +1,145 @@
...
+    throw err
+  }
+}
+
+const prepare = co.wrap(function * (target) {
+  if (target.host === 'localhost') {
+    exec(`createdb ${connstring(target, true)}`)
+  } else {
+    let result = yield psql.exec(target, 'SELECT COUNT(*) = 0 AS empty FROM pg_stat_user_tables')
+    if (!result.startsWith(' empty \n-------\n t')) throw new Error(`Remote database is not empty. Please create a new database or use ${cli.color.cmd('heroku pg:reset')}`)
```
msakrejda, line 36: Ugh but I'm guessing that comes from current behavior. Let's leave it and deal with it later if there are complaints.

commands/pull.js
```diff
@@ -0,0 +1,145 @@
...
+  // are used by the remote app, so take the final precaution of warning if
+  // the extensions available in the local database don't match. We don't
+  // report it if the difference is solely in the version of an extension
+  // used, though.
+  let sql = 'SELECT extname FROM pg_extension ORDER BY extname;'
+  let extensions = yield {
+    target: psql.exec(target, sql),
+    source: psql.exec(source, sql)
+  }
+  if (extensions.target !== extensions.source) {
```
msakrejda, line 57: I don't think we should care if there are _more_ extensions in the target than the source: I can't imagine that breaking anything. But again, maybe something to tackle in the future.

commands/pull.js
```diff
@@ -0,0 +1,145 @@
...
+HINT: You should review output to ensure that any errors
+ignored are acceptable - entire tables may have been missed, where a dependency
+could not be resolved. You may need to to install a postgresql-contrib package
+and retry.`)
+  }
+})
+
+const run = co.wrap(function * (source, target) {
+  yield prepare(target)
+  let password = p => p ? ` PGPASSWORD="${p}"` : ''
```
msakrejda, line 72: This doesn't handle passwords with double quote characters. We don't generate these, but users may be using that locally. I'm guessing this is another legacy thing copied though, we probably don't need to worry about it now.

lib/psql.js
```diff
@@ -11,11 +11,11 @@ function * exec (db, query) {
   debug(query)
   let env = Object.assign({}, process.env, {
     PGAPPNAME: 'psql non-interactive',
-    PGSSLMODE: 'require',
-    PGUSER: db.user,
+    PGSSLMODE: db.host === 'localhost' ? 'prefer' : 'require',
```
msakrejda: `db.hostname`, no?

lib/psql.js
```diff
@@ -11,11 +11,11 @@ function * exec (db, query) {
   debug(query)
   let env = Object.assign({}, process.env, {
```
msakrejda, line 12: We should let `process.env` override these values, not the other way around, no?

msakrejda (changes requested):
I think just the host -> hostname thing needs changing, the other stuff we can deal with as issues arise (although the `prepare` thing is broken on my machine).

lib/psql.js
jdx: I feel like users might have things like PGHOST set so they can easily connect to databases, so if we did that things would fail. @ransombriggs thoughts?

msakrejda: Oh I see. Well, if this is consistent with existing behavior, +1 on leaving it as is.

ransombriggs:
@dickeyxxx @uhoh-itsmaciek we should get rid of this pattern at this point, as a lot of these commands are totally broken on Windows because of it.  This is due to the fact that there is no `env` command, and no shorthand for environment variables.  Instead can we run the command with these variables injected into the environment.  Running inside of `git bash` generally does not help because if it needs to prompt for password it has issues because of tty issues.

```
env FOO=bar psql "some sql"
```

lib/psql.js
ransombriggs: @dickeyxxx @uhoh-itsmaciek yeah, I believe it is consistent, I would leave as is.

msakrejda:
@ransombriggs great point. Thoughts on how should we approach this? For something like `pg:pull` and `pg:push`, we basically have to interact with a locally-installed `pg_dump` and `pg_restore` one way or another, and they don't accept passwords on the command line (I think because they philosophically oppose an approach that could easily leave the password in the process listing--our usage of `env` does this anyway, but that's a separate issue). We could pass the option `--no-password` which will avoid prompting for a password (and maybe even detect and enrich the error message?) so maybe bypass some of the Windows tty issues--do you think that would help?

lib/psql.js
jdx: I think the old one used command line arguments but somehow I like env vars more (I think the end result would always be the same since flags override env vars anyways)

commands/pull.js
jdx: yeah this matches existing behavior. It does seem like a strange way to detect if the db is empty.

commands/pull.js
jdx: yeah good point, I think we should change this to just be less than

commands/pull.js
jdx: it's not quite equivalent, but I think it's better than what was before: https://github.com/heroku/heroku/blob/master/lib/heroku/helpers/pg_dump_restore.rb#L72

lib/psql.js
jdx: ah yes, updated

jdx:
yeah if you've got an idea to bypass the password I'd love to change this

commands/pull.js
msakrejda: That's tricky though because `.target` and `.source` are the full textual output of the query, and not an array or something :cry:.

commands/pull.js
msakrejda: It's not strange it's bad. But let's just :see_no_evil: here for now.

ransombriggs:
@uhoh-itsmaciek @dickeyxxx I assumed that you could do passwords through env vars like you do here, unless I am missing something.  https://github.com/heroku/heroku-pg/pull/43/files#diff-c88989f1047d0c743e308d3f556c5a59R12

msakrejda:
@ransombriggs @dickeyxxx that's not that simple because we pass off the whole pipeline to the shell (and that's probably a good idea--it's much easier than trying to wire together file descriptors in the calling code), so we can only set one `PGPASSWORD` for the pipeline, which can't cover both dump and restore.

ransombriggs:
@uhoh-itsmaciek thanks for reminding me about the piping together of the two commands.  Is there a way we can make this work in windows though, short of wiring together file descriptors?  I have seen several tickets around this, and would rather have this either fully work in windows or  error out indicating lack of support.  Wiring together the stdin / stdout I think will be more straightforward in JavaScript than Ruby and I think we could give it a shot.

msakrejda:
I'm not familiar with what is and isn't available in Windows, unfortunately. +1 on failing cleanly until we can handle that, though.

commands/pull.js
jdx: yeah it's definitely harder than I thought, I'll leave a TODO since this matches how it worked before

commands/pull.js
jdx: haha agreed

jdx:
what would you like to see fixed in this PR @uhoh-itsmaciek?

msakrejda:
Just the hostname thing, so ship it.

msakrejda (approved):

PR #44, opened 2016-10-11
added pg:diagnose

jdx:


commands/diagnose.js
```diff
@@ -0,0 +1,81 @@
...
+      body: params
+    })
+  })
+
+  let displayReport = report => {
+    cli.log(`Report ${report.id} for ${report.app}::${report.database}
+available for one month after creation on ${report.created_at}
+`)
+    let display = checks => {
+      if (!checks.length) return
```
msakrejda: Is this necessary? If `checks` is empty, the code won't execute the body of the loop and exit, right?

commands/diagnose.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    display(report.checks.filter(c => c.status === 'red'))
+    display(report.checks.filter(c => c.status === 'yellow'))
+    display(report.checks.filter(c => c.status === 'green'))
+    display(report.checks.filter(c => !['red', 'yellow', 'green'].find(d => d === c.status)))
+  }
+
+  let report
+  let id = args['DATABASE|REPORT_ID']
+  if (id && id.match(/^[a-z0-9\-]{36}$/)) {
```
msakrejda, line 59: You don't need the backslash here, I think: a trailing `-` in a matching group is not treated as a special character.

lib/util.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+  return sortBy(keys, k => k !== 'DATABASE_URL', 'name')
+}
+
+exports.databaseNameFromUrl = (uri, config) => {
+  const url = require('url')
+
+  let names = exports.configVarNamesFromValue(config, uri)
+  let name = names.pop()
+  while (name === 'DATABASE_URL') name = names.pop()
```
msakrejda: This should probably be the subject of another PR, but I noticed this behavior breaks when `DATABASE_URL` is the only attachment:

```
maciek@mothra:~$ h pg:info --app young-ridge-18903
=== 
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.5.4
Created:     2016-11-01 18:56 UTC
Data Size:   7.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported
Add-on:      postgresql-infinite-56908

maciek@mothra:~$ h addons --app young-ridge-18903

Add-on                                         Plan       Price  State  
─────────────────────────────────────────────  ─────────  ─────  ───────
heroku-postgresql (postgresql-infinite-56908)  hobby-dev  free   created
 └─ as DATABASE
```

(note the empty name after the `===` header in the `pg:info` output)

msakrejda:
+1

msakrejda (approved):

commands/diagnose.js
jdx: yeah I think it was before a refactor, removed

commands/diagnose.js
jdx: it is a special character, because you use it for declaring ranges of characters like `[a-zA-Z]+`

lib/util.js
jdx: ok so maybe this instead?

``` js
while (names.length > 0 && name === 'DATABASE_URL') name = names.pop()
```

commands/diagnose.js
msakrejda: Not as the last character in a character group expression. That is,

```
> "bar-".match(/[ac-]/g)
[ 'a', '-' ]
> "bar-".match(/[a-c]/g)
[ 'b', 'a' ]
> 
```

But I don't think this matters much; it can stay as is.

lib/util.js
uhoh-itsmaciek: `> 1`, right? The condition is evaluated before the loop body, so you want to make sure you have something to pop off.

commands/diagnose.js
jdx: ah yep, you're right

PR #45, opened 2016-11-16
Support bastions

idan:
/cc @heroku/dod-ux

Right now none of the heroku/heroku-pg-extras commands will work for private dbs because they don't know how to set up a tunnel through the bastion.

### DIFF
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -39,8 +39,7 @@ function buildTunnelPromise (db, dbTunnelConfig, timeout) {
       tunnel(dbTunnelConfig, (err, tnl) => {
         if (err) {
           debug(err)
-          cli.error(`Unable to establish a secure tunnel to your database.`)
-          process.exit(1)
+          reject(`Unable to establish a secure tunnel to your database.`)
         }
         debug('Tunnel created')
         resolve(tnl)
@@ -62,7 +61,7 @@ function buildPsqlPromise (query, dbEnv, timeout) {
     })
     psql.stderr.on('data', (data) => {
       debug(stripEOF(data.toString('utf8')))
-      cli.error(`Unable to execute the command.`)
+      reject(`Unable to execute the command.`)
       process.exit(1)
     })
     psql.on('error', (err) => {
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -85,7 +85,9 @@ function psqlInteractive (dbEnv, prompt, timeout) {
         reject(err)
       }
     })
-    resolve()
+    psql.on('close', (data) => {
+      resolve()
+    })
   })
 }
 
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -16,7 +16,7 @@ For help installing psql, see https://devcenter.heroku.com/articles/heroku-postg
 function env (db) {
   return Object.assign({}, process.env, {
     PGAPPNAME: 'psql non-interactive',
-    PGSSLMODE: db.hostname === 'localhost' ? 'prefer' : 'require',
+    PGSSLMODE: db.host === 'localhost' ? 'prefer' : 'require',
     PGUSER: db.user || '',
     PGPASSWORD: db.password,
     PGDATABASE: db.database,
@@ -64,7 +64,9 @@ function buildPsqlPromise(query, dbEnv) {
   const {spawn} = require('child_process')
   return new Promise((resolve, reject) => {
     let psql = spawn('psql', ['-c', query], {env: dbEnv, encoding: 'utf8'})
-    psql.stdout.on('data', (data) => (resolve(stripEOF(data))))
+    psql.stdout.on('data', (data) => {
+      resolve(stripEOF(data))
+    })
     psql.stderr.on('data', (data) => {
       debug(stripEOF(data.toString('utf8')))
       cli.error(`Unable to execute the command.`)
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -3,7 +3,6 @@
 const co = require('co')
 const debug = require('debug')('psql')
 const tunnel = require('tunnel-ssh')
-const cli = require('heroku-cli-util')
 
 function env (db) {
   return Object.assign({}, process.env, {
@@ -123,8 +122,6 @@ function * exec (db, query, timeout = 20000) {
   const command = tunnelPromise.then((tnl) => {
     let psqlPromise = execPsql(query, configs.dbEnv, timeout)
     return psqlPromise
-  }).catch((error) => {
-    cli.error(error)
   })
 
   return yield command
@@ -141,9 +138,6 @@ function * interactive (db) {
   const command = tunnelPromise.then((tnl) => {
     let psqlPromise = psqlInteractive(configs.dbEnv, prompt)
     return psqlPromise
-  }).catch((error) => {
-    cli.error(error)
-    process.exit(1)
   })
 
   return yield command
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -93,6 +93,7 @@ function * exec (db, query) {
 }
 
 function * interactive (db) {
+  //TODO: support tunneling in interactive as well
   const {spawnSync} = require('child_process')
   const pgUtil = require('./util')
   let name = pgUtil.getUrl(db.attachment.config_vars).replace(/^HEROKU_POSTGRESQL_/, '').replace(/_URL$/, '')
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -44,6 +44,7 @@ function sshTunnel (db, dbTunnelConfig, timeout) {
   return new Promise((resolve, reject) => {
     // if necessary to tunnel, setup a tunnel
     // see also https://github.com/heroku/heroku/blob/master/lib/heroku/helpers/heroku_postgresql.rb#L53-L80
+    let timer = setTimeout(() => reject('Establishing a secure tunnel timed out'), timeout)
     if (db.bastionKey) {
       tunnel(dbTunnelConfig, (err, tnl) => {
         if (err) {
@@ -51,28 +52,30 @@ function sshTunnel (db, dbTunnelConfig, timeout) {
           reject(`Unable to establish a secure tunnel to your database.`)
         }
         debug('Tunnel created')
+        clearTimeout(timer)
         resolve(tnl)
       })
     } else {
+      clearTimeout(timer)
       resolve()
     }
-    setTimeout(() => reject('Establishing a secure tunnel timed out'), timeout)
   })
 }
 
 function execPsql (query, dbEnv, timeout) {
   const {spawn} = require('child_process')
   return new Promise((resolve, reject) => {
+    let timer = setTimeout(() => reject('psql call timed out'), timeout)
     let result = ''
     let psql = spawn('psql', ['-c', query], {env: dbEnv, encoding: 'utf8', stdio: [ 'ignore', 'pipe', 'inherit' ]})
     psql.stdout.on('data', function (data) {
       result += data.toString()
     })
     psql.on('close', function (code) {
+      clearTimeout(timer)
       resolve(result)
     })
     handlePsqlError(reject, psql)
-    setTimeout(() => reject('psql call timed out'), timeout)
   })
 }
 
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -8,7 +8,7 @@ const cli = require('heroku-cli-util')
 function env (db) {
   return Object.assign({}, process.env, {
     PGAPPNAME: 'psql non-interactive',
-    PGSSLMODE: db.host === 'localhost' ? 'prefer' : 'require',
+    PGSSLMODE: db.hostname === 'localhost' ? 'prefer' : 'require',
     PGUSER: db.user || '',
     PGPASSWORD: db.password,
     PGDATABASE: db.database,
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -39,7 +39,7 @@ function tunnelConfig (db) {
   }
 }
 
-function buildTunnelPromise(db, dbTunnelConfig) {
+function buildTunnelPromise (db, dbTunnelConfig, timeout) {
   return new Promise((resolve, reject) => {
     // if necessary to tunnel, setup a tunnel
     // see also https://github.com/heroku/heroku/blob/master/lib/heroku/helpers/heroku_postgresql.rb#L53-L80
@@ -56,10 +56,13 @@ function buildTunnelPromise(db, dbTunnelConfig) {
     } else {
       resolve()
     }
+    setTimeout(() => reject('Establishing a secure tunnel timed out'), timeout)
+  }).catch((error) => {
+    console.log(error)
   })
 }
 
-function buildPsqlPromise(query, dbEnv) {
+function buildPsqlPromise (query, dbEnv, timeout) {
   const stripEOF = require('strip-eof')
   const {spawn} = require('child_process')
   return new Promise((resolve, reject) => {
@@ -76,10 +79,13 @@ function buildPsqlPromise(query, dbEnv) {
       debug(err)
       handleError(err)
     })
+    setTimeout(() => reject('psql call timed out'), timeout)
+  }).catch((error) => {
+    console.log(error)
   })
 }
 
-function buildPsqlInteractivePromise(dbEnv, prompt) {
+function buildPsqlInteractivePromise (dbEnv, prompt) {
   const {spawn} = require('child_process')
   return new Promise((resolve, reject) => {
     let psql = spawn('psql',
@@ -92,13 +98,13 @@ function buildPsqlInteractivePromise(dbEnv, prompt) {
   })
 }
 
-function getConfigs(db) {
+function getConfigs (db) {
   let dbEnv = env(db)
   const dbTunnelConfig = tunnelConfig(db)
   if (db.bastionKey) {
     dbEnv = Object.assign(dbEnv, {
       PGPORT: dbTunnelConfig.localPort,
-      PGHOST: dbTunnelConfig.localHost,
+      PGHOST: dbTunnelConfig.localHost
     })
   }
   return {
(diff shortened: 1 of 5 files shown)


idan:
I have to run, @camilleldn taking over

camillebaldock:
I have split https://github.com/heroku/heroku-pg/pull/46 (fixing the config var lookup bug for private databases) out as a separate pull request. This PR should focus on implementing SSH tunneling for databases behind bastions.

camillebaldock:
https://github.com/heroku/heroku-redis-jsplugin/blob/master/lib/commands/redis/cli.js#L110 for future reference

lib/psql.js
```diff
@@ -15,7 +16,7 @@ For help installing psql, see https://devcenter.heroku.com/articles/heroku-postg
 function env (db) {
   return Object.assign({}, process.env, {
     PGAPPNAME: 'psql non-interactive',
-    PGSSLMODE: db.hostname === 'localhost' ? 'prefer' : 'require',
+    PGSSLMODE: db.host === 'localhost' ? 'prefer' : 'require',
```
msakrejda: Shouldn't this be `hostname`? `host` includes the port if one is set explicitly, which should be moot for the purpose of this conditional.

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+function buildTunnelPromise (db, dbTunnelConfig, timeout) {
+  return new Promise((resolve, reject) => {
+    // if necessary to tunnel, setup a tunnel
+    // see also https://github.com/heroku/heroku/blob/master/lib/heroku/helpers/heroku_postgresql.rb#L53-L80
+    if (db.bastionKey) {
+      tunnel(dbTunnelConfig, (err, tnl) => {
+        if (err) {
+          debug(err)
+          cli.error(`Unable to establish a secure tunnel to your database.`)
+          process.exit(1)
```
msakrejda: The `heroku-cli` convention is to throw an error rather than `cli.error` and exit.

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+        }
+        debug('Tunnel created')
+        resolve(tnl)
+      })
+    } else {
+      resolve()
+    }
+    setTimeout(() => reject('Establishing a secure tunnel timed out'), timeout)
+  }).catch((error) => {
+    cli.error(error)
```
msakrejda: I think you can drop this `catch`, no? You want to propagate the error anyway to ensure the program terminates (since establishing the tunnel is necessary for this to work).

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+  })
+}
+
+function buildPsqlPromise (query, dbEnv, timeout) {
+  const stripEOF = require('strip-eof')
+  const {spawn} = require('child_process')
+  return new Promise((resolve, reject) => {
+    let psql = spawn('psql', ['-c', query], {env: dbEnv, encoding: 'utf8'})
+    psql.stdout.on('data', (data) => {
+      resolve(stripEOF(data))
```
msakrejda: Can `stdout` produce a 'data' event multiple times? What happens in that case? Maybe we want to accumulate and then concatenate the produced data and resolve the promise on process exit?

lib/psql.js
jdx: shouldn't this be `hostname`?

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+    })
+    psql.stderr.on('data', (data) => {
+      debug(stripEOF(data.toString('utf8')))
+      cli.error(`Unable to execute the command.`)
+      process.exit(1)
```
msakrejda: I don't think that anything on stderr is automatically an error...

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
-  const {spawnSync} = require('child_process')
+function * exec (db, query, timeout = 20000) {
   handleSignals()
   debug(query)
-  let {stdout, error: err, status} = spawnSync('psql', ['--command', query], {env: env(db), encoding: 'utf8', stdio: [0, 'pipe', 2]})
-  handleError(err)
-  if (status !== 0) process.exit(status)
-  return stripEOF(stdout)
+  let configs = getConfigs(db)
+  const tunnelPromise = buildTunnelPromise(db, configs.dbTunnelConfig, timeout)
```
jdx: This section could be more direct:

```js
yield buildTunnelPromise(db, configs.dbTunnelConfig, timeout)
return yield buildPsqlPromise(query, configs.dbEnv, timeout)
```

Also the names of these functions should probably be something more like `tunnel()` or `createTunnel()` and `execPSQL()` or `psql()`. The fact it's generating a promise isn't as important as what the function does.

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+      resolve(stripEOF(data))
+    })
+    psql.stderr.on('data', (data) => {
+      debug(stripEOF(data.toString('utf8')))
+      cli.error(`Unable to execute the command.`)
+      process.exit(1)
+    })
+    psql.on('error', (err) => {
+      debug(err)
+      handleError(err)
```
msakrejda: This uses the same handler that tells you you don't seem to have `psql` installed which isn't really true. Maybe just `reject`?

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+    })
+    setTimeout(() => reject('psql call timed out'), timeout)
```
msakrejda: Nice.

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+    cli.error(error)
+  })
+}
+
+function buildPsqlPromise (query, dbEnv, timeout) {
+  const stripEOF = require('strip-eof')
+  const {spawn} = require('child_process')
+  return new Promise((resolve, reject) => {
+    let psql = spawn('psql', ['-c', query], {env: dbEnv, encoding: 'utf8'})
+    psql.stdout.on('data', (data) => {
```
jdx: I worry this might not work if the output is large. Maybe we could test it with the `pg:mandelbrot` repeated a few times? I can help with that if needed

lib/psql.js
msakrejda: Although I guess you're in a promise here so maybe `reject` the promise instead? I suppose if you want to definitely exit immediately you can do this, but it feels weird to have that buried in here.

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+    localPort: localPort
+  }
+}
+
+function buildTunnelPromise (db, dbTunnelConfig, timeout) {
+  return new Promise((resolve, reject) => {
+    // if necessary to tunnel, setup a tunnel
+    // see also https://github.com/heroku/heroku/blob/master/lib/heroku/helpers/heroku_postgresql.rb#L53-L80
+    if (db.bastionKey) {
+      tunnel(dbTunnelConfig, (err, tnl) => {
```
jdx, line 49: it would be nice if we had test coverage over this section

jdx (changes requested):

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+      cli.error(`Unable to execute the command.`)
+      process.exit(1)
+    })
+    psql.on('error', (err) => {
+      debug(err)
+      handleError(err)
+    })
+    setTimeout(() => reject('psql call timed out'), timeout)
+  }).catch((error) => {
+    cli.error(error)
```
msakrejda: Same here, why not just `reject`?

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+
+function buildPsqlInteractivePromise (dbEnv, prompt) {
+  const {spawn} = require('child_process')
+  return new Promise((resolve, reject) => {
+    let psql = spawn('psql',
+                     ['--set', `PROMPT1=${prompt}`, '--set', `PROMPT2=${prompt}`],
+                     {env: dbEnv, stdio: 'inherit'})
+    psql.on('error', (err) => {
+      debug(err)
+      handleError(err)
```
msakrejda: Same as above, I think `reject`-ing would be clearer here, no?

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+    })
+    setTimeout(() => reject('psql call timed out'), timeout)
+  }).catch((error) => {
+    cli.error(error)
+  })
+}
+
+function buildPsqlInteractivePromise (dbEnv, prompt) {
+  const {spawn} = require('child_process')
+  return new Promise((resolve, reject) => {
```
jdx, line 84: this is creating a promise that's never resolved

lib/psql.js
```diff
@@ -24,31 +25,123 @@ function env (db) {
...
+function * exec (db, query, timeout = 20000) {
   handleSignals()
   debug(query)
-  let {stdout, error: err, status} = spawnSync('psql', ['--command', query], {env: env(db), encoding: 'utf8', stdio: [0, 'pipe', 2]})
-  handleError(err)
-  if (status !== 0) process.exit(status)
-  return stripEOF(stdout)
+  let configs = getConfigs(db)
+  const tunnelPromise = buildTunnelPromise(db, configs.dbTunnelConfig, timeout)
+  const command = tunnelPromise.then((tnl) => {
```
msakrejda: Also the explicit `promise` naming seems a little idiosyncratic: most JS I've seen handles that implicitly (e.g., maybe something like

```javascript
const tunnel = buildTunnel(...)
const command = tunnel.then((tnl) => runPsql(query, ...)
```

or something). I think the specific naming and formatting is not that important, but referencing promises explicitly feels clunky.

msakrejda (changes requested):
Mostly looks good, although I think we need to tweak stdout handling to make sure we get it all (if, e.g., there are multiple "chunks"). Other than that just some style comments.

msakrejda:
@dickeyxxx may also want to take a look...

msakrejda:
Ha @dickeyxxx has already had a look. Thanks!

lib/psql.js
camillebaldock: correct, it just hangs on massive query results. i'm fixing

lib/psql.js
camillebaldock: done :)

camillebaldock:
@uhoh-itsmaciek @dickeyxxx addressed a bunch of your comments, thanks again for your review ❤️ , I would appreciate a final look now that the code has changed :)

lib/psql.js
```diff
@@ -24,32 +17,136 @@ function env (db) {
...
+  })
+}
+
+function psqlInteractive (dbEnv, prompt, timeout) {
+  const {spawn} = require('child_process')
+  return new Promise((resolve, reject) => {
+    let psql = spawn('psql',
+                     ['--set', `PROMPT1=${prompt}`, '--set', `PROMPT2=${prompt}`],
+                     {env: dbEnv, stdio: 'inherit'})
+    psql.on('error', (err) => {
```
jdx: seems like this function could be shared with line 71

lib/psql.js
```diff
@@ -24,32 +17,136 @@ function env (db) {
...
-  if (status !== 0) process.exit(status)
+  let configs = getConfigs(db)
+
+  const tunnelPromise = sshTunnel(db, configs.dbTunnelConfig)
+  const command = tunnelPromise.then((tnl) => {
+    let psqlPromise = psqlInteractive(configs.dbEnv, prompt)
+    return psqlPromise
+  }).catch((error) => {
+    cli.error(error)
+    process.exit(1)
```
jdx: ideally it should fail with the same status code that `psql` failed with, probably the same in non-interactive

lib/psql.js
```diff
@@ -24,32 +17,136 @@ function env (db) {
...
+  return new Promise((resolve, reject) => {
+    let result = ''
+    let psql = spawn('psql', ['-c', query], {env: dbEnv, encoding: 'utf8'})
+    psql.stdout.on('data', function (data) {
+      result += data.toString()
+    })
+    psql.on('close', function (code) {
+      resolve(result)
+    })
+    psql.stderr.on('data', (data) => {
```
jdx: we should just propagate stderr out to the terminal like we did before. Just because there is something on stderr doesn't mean it's an error either.

lib/psql.js
```diff
@@ -24,32 +17,136 @@ function env (db) {
...
-  handleError(err)
-  if (status !== 0) process.exit(status)
-  return stripEOF(stdout)
+  let configs = getConfigs(db)
+
+  const tunnelPromise = sshTunnel(db, configs.dbTunnelConfig, timeout)
+  const command = tunnelPromise.then((tnl) => {
+    let psqlPromise = execPsql(query, configs.dbEnv, timeout)
+    return psqlPromise
+  }).catch((error) => {
```
jdx: no need to catch this, just let the error propagate

lib/psql.js
```diff
@@ -24,32 +17,136 @@ function env (db) {
...
+function * exec (db, query, timeout = 20000) {
   handleSignals()
-  debug(query)
-  let {stdout, error: err, status} = spawnSync('psql', ['--command', query], {env: env(db), encoding: 'utf8', stdio: [0, 'pipe', 2]})
-  handleError(err)
-  if (status !== 0) process.exit(status)
-  return stripEOF(stdout)
+  let configs = getConfigs(db)
+
+  const tunnelPromise = sshTunnel(db, configs.dbTunnelConfig, timeout)
```
jdx: I would like to see these lines use generators instead:

```js
yield sshTunnel(db, configs.dbTunnelConfig, timeout)
yield execPsql(query, configs.dbEnv, timeout)
```

jdx (changes requested):
looks pretty good, a couple more things I'd like to see fixed though

lib/psql.js
msakrejda: Yeah, that makes sense. So `stdio: [ 'ignore', 'pipe', 'inherit' ]` in `spawn` above, right?

lib/psql.js
```diff
@@ -24,32 +17,136 @@ function env (db) {
...
+      resolve(result)
+    })
+    psql.stderr.on('data', (data) => {
+      debug(stripEOF(data.toString('utf8')))
+      reject(`Unable to execute the command.`)
+      process.exit(1)
+    })
+    psql.on('error', (err) => {
+      if (err.code === 'ENOENT') {
+        reject(`The local psql command could not be located. For help installing psql, see https://devcenter.heroku.com/articles/heroku-postgresql#local-setup`)
```
msakrejda: Nice

lib/psql.js
msakrejda: Yeah, if you just bring it up to top level, you can share it pretty easily.

lib/psql.js
```diff
@@ -24,32 +17,136 @@ function env (db) {
...
+                     ['--set', `PROMPT1=${prompt}`, '--set', `PROMPT2=${prompt}`],
+                     {env: dbEnv, stdio: 'inherit'})
+    psql.on('error', (err) => {
+      if (err.code === 'ENOENT') {
+        reject(`The local psql command could not be located. For help installing psql, see https://devcenter.heroku.com/articles/heroku-postgresql#local-setup`)
+      } else {
+        reject(err)
+      }
+    })
+    resolve()
```
msakrejda: Don't you want to wait for the program to exit (`.on('close')`) to resolve this?

lib/psql.js
jdx: yes, and we could probably pipe in stdin as well, though I'm not sure if psql would ever use it

msakrejda: I think it ignores it if you pass `-c`:

```
maciek@mothra:~$ <<<'select 2' psql -qc 'select 1'
 ?column? 
----------
        1
(1 row)
```

lib/psql.js
```diff
@@ -24,32 +16,115 @@ function env (db) {
...
+    localPort: localPort
+  }
+}
+
+function handlePsqlError (reject, psql) {
+  psql.on('error', (err) => {
+    if (err.code === 'ENOENT') {
+      reject(`The local psql command could not be located. For help installing psql, see https://devcenter.heroku.com/articles/heroku-postgresql#local-setup`)
+    } else {
+      Promise.reject(err)
```
msakrejda: Shouldn't this also be `reject(err)`?

lib/psql.js
```diff
@@ -24,32 +16,115 @@ function env (db) {
...
+          debug(err)
+          reject(`Unable to establish a secure tunnel to your database.`)
+        }
+        debug('Tunnel created')
+        resolve(tnl)
+      })
+    } else {
+      resolve()
+    }
+    setTimeout(() => reject('Establishing a secure tunnel timed out'), timeout)
```
msakrejda: Oh ugh I just realized that if you do this I think you need to cancel the timeout (with the token from the return value of `setTimeout`) when the tunnel does resolve successfully (or isn't needed). I think it's also fine to skip the timeout instead.

lib/psql.js
```diff
@@ -24,32 +16,115 @@ function env (db) {
...
+    let result = ''
+    let psql = spawn('psql', ['-c', query], {env: dbEnv, encoding: 'utf8', stdio: [ 'ignore', 'pipe', 'inherit' ]})
+    psql.stdout.on('data', function (data) {
+      result += data.toString()
+    })
+    psql.on('close', function (code) {
+      resolve(result)
+    })
+    handlePsqlError(reject, psql)
+    setTimeout(() => reject('psql call timed out'), timeout)
```
msakrejda: Same as above re: timeout.

msakrejda:
Ok, some tiny comments left (I'd drop the `setTimeouts` for now if you don't feel like adding the cancellation--I don't think that's that'd big a deal).

msakrejda:
+1 :tada:

msakrejda (approved):

camillebaldock:
@dickeyxxx any last feedback? :)

test/lib/psql.js
```diff
@@ -3,36 +3,93 @@
 /* global describe it */
 
 const sinon = require('sinon')
-const psql = require('../../lib/psql')
-
+const Code = require('code')
```
jdx: can we use `unexpected` instead since it's already used by this project?

jdx (changes requested):

test/lib/psql.js
camillebaldock: fixed :)

package.json
```diff
@@ -22,9 +22,11 @@
     "sinon": "1.17.6",
     "smooth-progress": "1.1.0",
     "string": "3.3.3",
-    "strip-eof": "1.0.0"
+    "strip-eof": "1.0.0",
+    "tunnel-ssh": "4.1.1"
   },
   "devDependencies": {
+    "code": "^4.0.0",
```
jdx: We don't need this now

jdx (approved):

PR #46, opened 2016-11-16
Support private DBs in database fetcher

camillebaldock:
Currently, all plugins such as heroku-pg-extras fail on fetching the database payload (https://github.com/heroku/heroku-pg-extras/issues/137). 

This is because we cannot rely on `addon.config_vars[0]` being the database url in the private database case because of the presence of bastion URLs in the config vars.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,0 +1,56 @@
+const debug = require('./debug')
+const url = require('url')
+const sample = require('lodash.sample')
+
+const getBastion = function (config, baseName) {
+  // If there are bastions, extract a host and a key
+  // otherwise, return an empty Object
+
+  // If there are bastions:
+  // * there should be one *_BASTION_KEY
+  // * pick one host from the comma-separated list in *_BASTIONS
+  // We assert that _BASTIONS and _BASTION_KEY always exist together
+  // If either is falsy, pretend neither exist
+
+  const bastionKey = config[`${baseName}_BASTION_KEY`]
+  const bastionHost = sample(config[`${baseName}_BASTIONS`].split(','))
+  return (!(bastionKey && bastionHost))
+    ? {}
+    : {bastionHost, bastionKey}
+}
+
+exports.getConnectionDetails = function (addon, config) {
+  const connstrings = addon.config_vars
+    .filter((cv) => (
+      config[cv].startsWith('postgres://')
+      && cv.endsWith('_URL')
+    ))
+
+  if (connstrings.length === 0) {
+    // TODO blow up in a useful fashion
+  }
+
+  const baseName = connstrings[0].slice(0, -4)
+
+  // build the default payload for non-bastion dbs
+  debug(`Using "${connstrings[0]}" to connect to your database…`)
+  const target = url.parse(config[connstrings[0]])
+  let [user, password] = target.auth.split(':')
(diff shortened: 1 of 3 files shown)


lib/util.js
```diff
@@ -0,0 +1,56 @@
...
+exports.getConnectionDetails = function (addon, config) {
+  const connstrings = addon.config_vars
+    .filter((cv) => (
+      config[cv].startsWith('postgres://') &&
+      cv.endsWith('_URL')
+    ))
+
+  if (connstrings.length === 0) {
+    const cli = require('heroku-cli-util')
+    cli.error('Database URL not found for this addon')
```
jdx: `cli.error()` (which is maybe badly named) doesn't actually error out. If you want to do that, I would just raise an exception:

``` js
throw new Error('Database URL not found for this addon')
```

jdx (changes requested):

lib/util.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const bastionKey = config[`${baseName}_BASTION_KEY`]
+  const bastionHost = sample((config[`${baseName}_BASTIONS`] || '').split(','))
+  return (!(bastionKey && bastionHost))
+    ? {}
+    : {bastionHost, bastionKey}
+}
+
+exports.getConnectionDetails = function (addon, config) {
+  const connstrings = addon.config_vars
```
msakrejda: Something like `connstringVars` would be clearer here...

lib/util.js
```diff
@@ -0,0 +1,56 @@
...
+      config[cv].startsWith('postgres://') &&
+      cv.endsWith('_URL')
+    ))
+
+  if (connstrings.length === 0) {
+    const cli = require('heroku-cli-util')
+    cli.error('Database URL not found for this addon')
+  }
+
+  const baseName = connstrings[0].slice(0, -4)
```
msakrejda: This is kind of opaque and combined with the confusing name it took me a moment to figure out what's going on. Given that this is not performance-sensitive code, maybe `connstringVars[0].replace(/_URL$/, '')` or something like that?

lib/util.js
```diff
@@ -0,0 +1,56 @@
...
+
+  // build the default payload for non-bastion dbs
+  debug(`Using "${connstrings[0]}" to connect to your database…`)
+  const target = url.parse(config[connstrings[0]])
```
msakrejda: For what it's worth, my rule of thumb is if you use an expression more than twice, create a local variable to alias it. It reads better and is easier to update.

msakrejda (changes requested):

msakrejda:
Jeff's point is the only blocker here.

lib/util.js
idan: Given the choice between adding an expository comment and adding a regex, I'd vote for adding a comment

lib/util.js
idan: +1

lib/util.js
msakrejda: That works too. I suggested the regex because it's a simple one and the code is self-documenting, but I'm fine with a comment.

msakrejda (approved):

jdx (approved):

PR #49, opened 2016-11-17
Workaround for Postgres 9.6, since it doesn't have waiting column

keiko713:
Tested locally, it... works.

### DIFF
--- a/commands/ps.js
+++ b/commands/ps.js
@@ -13,7 +13,7 @@ function * run (context, heroku) {
 
   let db = yield fetcher.database(app, database)
 
-  let waiting_query = `
+  let waitingQuery = `
 SELECT EXISTS(
   SELECT 1 FROM information_schema.columns WHERE table_schema = 'pg_catalog'
     AND table_name = 'pg_stat_activity'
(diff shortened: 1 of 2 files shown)


jdx (approved):

keiko713:
OH FORGOT TO REWRITE TEST, gimme a sec

commands/ps.js
```diff
@@ -13,13 +13,28 @@ function * run (context, heroku) {
...
+  let waiting_query = `
+SELECT EXISTS(
+  SELECT 1 FROM information_schema.columns WHERE table_schema = 'pg_catalog'
+    AND table_name = 'pg_stat_activity'
+    AND column_name = 'waiting'
+) AS available
+`
+
+  let waiting_output = yield psql.exec(db, waiting_query)
+  let waiting = 'CASE WHEN wait_event IS NULL THEN false ELSE true END AS waiting'
```
msakrejda: How about `wait_event IS NOT NULL AS waiting`?

commands/ps.js
```diff
@@ -13,13 +13,28 @@ function * run (context, heroku) {
...
+
+  if (waiting_output.includes('t')) {
```
msakrejda: This is error-prone because `.psqlrc` can include stuff that will send a `t` to the output, but we're making that same mistake elsewhere so I think that's okay for now.

msakrejda (changes requested):

commands/ps.js
keiko713: yeah I don't particularly like this but found similar thing here:
https://github.com/heroku/heroku-pg-extras/blob/07cda801f3368676afd54b6fee8b2874f3487701/lib/util.js#L7-L14

happy to improve if there is simple & nice way. any thoughts?

msakrejda: I don't think there is; let's stick with this for now.

keiko713: +1, at least easy to search when we replace everything

msakrejda (approved):

PR #56, opened 2016-11-22
Fix ambiguity errors & implicit attachments

ransombriggs:
@dickeyxxx could you review?

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,0 +1,9 @@
+'use strict'
+
+const memoizePromise = require('./memoize_promise')
+
+function config (heroku, app) {
+  return heroku.get(`/apps/${app}/config-vars`)
+}
+
+module.exports = memoizePromise(config, (_heroku, app) => app)
--- a/lib/fetcher.js
+++ b/lib/fetcher.js
@@ -7,16 +7,14 @@ const getConfig = require('./config')
 
 module.exports = heroku => {
   function * attachment (app, db) {
-    // this is kind of gross, but cannot do a generator in a catch
-    // so basically this makes it easier to handle the 422 & 404
-    function * matchesHelper (app, db) {
+    function matchesHelper (app, db) {
       const {resolve} = require('heroku-cli-addons')
 
       db = db || 'DATABASE_URL'
       debug(`fetching ${db} on ${app}`)
 
-      return yield resolve.appAttachment(heroku, app, db, {addon_service: 'heroku-postgresql'})
-      .then(attached => ({matches: [attached], err: null}))
+      return resolve.appAttachment(heroku, app, db, {addon_service: 'heroku-postgresql'})
+      .then(attached => ({matches: [attached]}))
       .catch(function (err) {
         if (err.statusCode === 422 && err.body && err.body.id === 'multiple_matches' && err.matches) {
           return {matches: err.matches, err: err}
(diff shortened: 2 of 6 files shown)


lib/config.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+const memoizePromise = require('./memoize_promise')
+
+function config (heroku, app) {
+  return heroku.get(`/apps/${app}/config-vars`)
+}
+
+module.exports = memoizePromise(config, (_heroku, app) => app)
```
jdx, line 9: nice

lib/fetcher.js
```diff
@@ -3,38 +3,99 @@
...
 const debug = require('./debug')
 const pgUtil = require('./util')
+const getConfig = require('./config')
 
 module.exports = heroku => {
   function * attachment (app, db) {
-    const {resolve} = require('heroku-cli-addons')
+    // this is kind of gross, but cannot do a generator in a catch
+    // so basically this makes it easier to handle the 422 & 404
+    function * matchesHelper (app, db) {
```
jdx: while this works, you should wrap it in `co.wrap()` so it's a regular promise

lib/fetcher.js
```diff
@@ -3,38 +3,99 @@
...
+      const {resolve} = require('heroku-cli-addons')
 
-    db = db || 'DATABASE_URL'
-    debug(`fetching ${db} on ${app}`)
-    return yield resolve.appAttachment(heroku, app, db, {addon_service: 'heroku-postgresql'})
+      db = db || 'DATABASE_URL'
+      debug(`fetching ${db} on ${app}`)
+
+      return yield resolve.appAttachment(heroku, app, db, {addon_service: 'heroku-postgresql'})
+      .then(attached => ({matches: [attached], err: null}))
```
jdx: shouldn't be necessary to pass the null err here

lib/fetcher.js
```diff
@@ -3,38 +3,99 @@
...
+
+        if (err.statusCode === 404 && err.body && err.body.id === 'not_found') {
+          return {matches: null, err: err}
+        }
+
+        throw err
+      })
+    }
+
+    let {matches, err} = yield matchesHelper(app, db)
```
jdx, line 32: I love this pattern. Multiple return objects is super useful.

lib/memoize_promise.js
```diff
@@ -0,0 +1,26 @@
...
+    const result = func.apply(this, args)
+
+    result.then(function () {
+      memoized.cache = cache.set(key, result) || cache
+      return arguments
+    })
+
+    return result
+  }
+  memoized.cache = new memoize.Cache()
```
jdx: do we really need to use `lodash.memoize`? I think if you replaced this line with `memoized.cache = new WeakMap()` it would be equivalent.

jdx (changes requested):

ransombriggs:
@dickeyxxx made changes in response to feedback in https://github.com/heroku/heroku-pg/pull/56/commits/6dd7be80dc16b2f24a1630f44ba6e15999b57766

ransombriggs:
@dickeyxxx I pushed up two more commits to fix bugs I found, I think this is ready for @uhoh-itsmaciek after your review.

jdx (approved):

PR #63, opened 2016-12-01
always catch sigint

### LINKED ISSUE #62: psql history not saved line-by-line
I'm not sure if this is a change or not, but it seems that if you run a set of commands and abort the pg:psql session via ctrl-c, the history isn't saved.

I thought that under the legacy pg:psql that history was saved each time you executed a statement, but I might be wrong.

jdx:
Fixes #62

### DIFF
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -108,7 +108,7 @@ function getConfigs (db) {
 }
 
 function handleSignals () {
-  process.once('SIGINT', () => {})
+  process.on('SIGINT', () => {})
 }
 
 function * exec (db, query, timeout = 20000) {


msakrejda:
Ha. I was looking at how you'd done signal handling, and saw this but then didn't see a `process.once` in the node docs and meant to ask you about it but forgot. Don't you need to explicitly propagate the signal to the psql child process, though? That's [how I got it working in heroku-run-local](https://github.com/uhoh-itsmaciek/heroku-run-localjs/commit/19b4b045dca11d8a69b44d8a6c2adf21f629c5bb).

jdx:
it doesn't seem that it needs to propagate... though I'm not sure why. But with this in if you run

```
heroku pg:psql -c "SELECT pg_sleep(1000000)"
--> Connecting to postgresql-perpendicular-14473
^CCancel request sent
ERROR:  canceling statement due to user request
```

it seems to work correctly

msakrejda:
Huh. Maybe signals propagate to children automagically? Anyway, tried it here and you're right, it seems to work. +1

msakrejda (approved):

jdx:
I think it might be because stdin is piped in and the signals pass through that?

PR #64, opened 2016-12-09
Wait on correct app

msakrejda:
Copies can be associated with the source or target, depending on which
one is actually a heroku-postgresql add-on. Right now, we assume that
the context app is the one to wait on, but that's not always the case.
Overload wait to allow waiting for the correct thing.

@keiko713 can you try this branch and see if it fixes the problem?

### DIFF
--- a/lib/pgbackups.js
+++ b/lib/pgbackups.js
@@ -64,7 +64,10 @@ module.exports = (context, heroku) => ({
       }
     }
   },
-  wait: (action, transferID, interval, verbose) => {
+  wait: (action, transferID, interval, verbose, app) => {
+    if (app === undefined) {
+      app = context.app
+    }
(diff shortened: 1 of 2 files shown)


keiko713:
I tested and this branch worked well!

lib/pgbackups.js
```diff
@@ -64,7 +64,10 @@ module.exports = (context, heroku) => ({
       }
     }
   },
-  wait: (action, transferID, interval, verbose) => {
+  wait: (action, transferID, interval, verbose, app) => {
```
msakrejda, line 67: I admit this is pretty gross. I welcome suggestions for a better approach, but this does seem to work.

jdx:
doh!

PR #66, opened 2016-12-10
fix plan name for pgdiagnose

### LINKED ISSUE #65: pg:diagnose is receiving the wrong format for plan name
pgdiagnose is expecting to receive the plan name like `standard-0`, instead of `heroku-postgresql:standard-0`.
This is causing the issue that we don't know the proper plan in pgdiagnose side, and always reporting RED for "Connection Count". The easy fix would be changing heroku-pg and pass the param that pgdiagnose expects.

https://github.com/heroku/pgdiagnose/blob/master/plans.go
https://github.com/heroku/heroku-pg/blob/master/test/commands/diagnose.js#L11

Perviously:

https://github.com/heroku/heroku/blob/master/lib/heroku/helpers/pg_diagnose.rb#L50

jdx:
Fixes #65

### DIFF
--- a/commands/diagnose.js
+++ b/commands/diagnose.js
@@ -16,7 +16,7 @@ function * run (context, heroku) {
     let config = yield heroku.get(`/apps/${app}/config-vars`)
     let params = {
       url: config[util.getUrl(db.config_vars)],
-      plan: db.plan.name,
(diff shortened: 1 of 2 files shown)


keiko713:
lgtm!

keiko713 (approved):

PR #70, opened 2016-12-13
Use process.env.PGPORT for pull / push if set

ransombriggs:
@uhoh-itsmaciek @keiko713 could you review?


msakrejda:
Looks legit. We should probably consider supporting others like PGPASSWORD, PGUSER, and PGDATABASE, but maybe we can wait until that comes up.

msakrejda (approved):

PR #74, opened 2016-12-14
Use bastions to tunnel in pg:pull & pg:push

ransombriggs:
@dickeyxxx was getting really close, but it still hangs.  See my debugging notes here, it appears that the tunnel is getting set up correctly, but that the command itself hangs for unknown reasons.  If you want the command we shell out and run it manually during that wait period it does return the expected output, not sure why it does not work through exec yet while psql seems to have no issues (although psql uses spawn and sets things up via environment variables passed down to spawn)

[1ca782669f9fa62a1c35d472b142325f55bda222](https://github.com/heroku/heroku-pg/commit/1ca782669f9fa62a1c35d472b142325f55bda222#diff-caacc9db9d4fdff25c7eacfa9b8d4fdcR94)

### DIFF
--- a/commands/pull.js
+++ b/commands/pull.js
@@ -5,6 +5,7 @@ const co = require('co')
 const debug = require('debug')('push')
 const psql = require('../lib/psql')
 const env = require('process').env
+const bastion = require('../lib/bastion')
 
 function parseURL (db) {
   const url = require('url')
@@ -69,11 +70,37 @@ and retry.`)
   }
 })
 
+const maybeTunnel = function * (herokuDb) {
+  const configs = bastion.getConfigs(herokuDb)
+  const tunnel = yield bastion.sshTunnel(herokuDb, configs.dbTunnelConfig)
+  if (tunnel) {
+    const tunnelHost = {
+      host: 'localhost',
+      port: configs.dbTunnelConfig.localPort
+    }
+
+    herokuDb = Object.assign(herokuDb, tunnelHost)
+  }
+  return herokuDb
+}
+
 const run = co.wrap(function * (source, target) {
   yield prepare(target)
+
   let password = p => p ? ` PGPASSWORD="${p}"` : ''
   let dump = `env${password(source.password)} PGSSLMODE=prefer pg_dump --verbose -F c -Z 0 ${connstring(source, true)}`
   let restore = `env${password(target.password)} pg_restore --verbose --no-acl --no-owner ${connstring(target)}`
+
+  // stuck here, take the output of this console.log
+  // and paste it in another windows and you can see it work
+  // but when run through exec it fails
+
+  // uncomment these lines to play around
+
+  // const wait = require('co-wait')
+  // console.log(dump)
+  // yield wait(100000)
+
   exec(`${dump} | ${restore}`)
   yield verifyExtensionsMatch(source, target)
 })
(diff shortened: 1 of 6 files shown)


msakrejda:
@ransombriggs I think this is the same thing that @camilleldn and @idan and I ran into when working on tunneling for the other commands: you can't `execSync` through the tunnel: the tunnel starts a server and needs to accept a local socket connection (from pg_dump in this case) and proxy it through the bastion, and `execSync` blocks until the command returns. So the problem is we

 1. Set up the tunnel
 2. `execSync` pg_dump
 3. pg_dump connects to the local socket the tunnel just opened and initiates a handshake
 4. The node process can't invoke the server new connection callback handler, because it's busy waiting for the `execSync` to return (and, of course, node is single-threaded)
 5. pg_dump is waiting for the handshake, and node is waiting for `execSync` to return

Essentially, `execSync` through a tunnel will always deadlock.

I solved this by using an [async `spawn` instead](https://github.com/heroku/heroku-pg/blob/a33f1472bf3039b21b504b4b8abec47a411749c2/lib/psql.js#L65-L80).

ransombriggs:
@uhoh-itsmaciek thanks, that was totally it, I pushed up https://github.com/heroku/heroku-pg/pull/74/commits/0bbeae349b5a0bca82d326921ccc2a728e7138fd with just that fix

commands/pull.js
```diff
@@ -30,6 +31,20 @@ function exec (cmd, opts = {}) {
...
 
+function spawn (cmd) {
+  const {spawn} = require('child_process')
+  return new Promise((resolve, reject) => {
+    let result = ''
+    let psql = spawn(cmd, [], {encoding: 'utf8', stdio: [ 'ignore', 'pipe', 'inherit' ], shell: true})
+    psql.stdout.on('data', function (data) {
+      result += data.toString()
+    })
+    psql.on('close', function (code) {
```
jdx, line 42: are we not interested in nonzero?

ransombriggs: @dickeyxxx I checked the old ruby code and they do not do anything with it so I kept the same behavior.  One thing I did notice was that the old one exits with `1` if the extensions do not match, which is something we do not do.

https://github.com/heroku/legacy-cli/blob/master/lib/heroku/helpers/pg_dump_restore.rb#L118
https://github.com/heroku/legacy-cli/blob/master/lib/heroku/helpers/pg_dump_restore.rb#L98

jdx: we do have that code though

jdx: https://github.com/heroku/heroku-pg/blob/4053f7dd92158b2af47445a51f125c908674c4b4/commands/pull.js#L47

jdx: personally I think it would be good to fail on nonzero even if the old CLI did not

ransombriggs: @dickeyxxx we have `verifyExtensionsMatch` but it does not exit non-zero, it just outputs a message to stderr

ransombriggs: @dickeyxxx I added an exit for non-zero codes in https://github.com/heroku/heroku-pg/pull/74/commits/d4b9c633041ca66bbcfa8bdc617c3414b32e5092 did you want me to change the `verifyExtensionsMatch` behavior or leave it alone?

jdx: oh that's maybe ok. Not having matching extensions seems like more of a warning scenario than an error one. Especially since it errors on non-equal extensions, but if there are more extensions on the target that should be fine.

jdx (approved):

PR #76, opened 2016-12-15
Pass through verbose to GET & check logs defined

ransombriggs:
@keiko713 can you review?

### DIFF
--- a/lib/pgbackups.js
+++ b/lib/pgbackups.js
@@ -86,9 +86,14 @@ module.exports = (context, heroku) => ({
       let backup
       let failures = 0
 
+      let url = `/client/v11/apps/${app}/transfers/${transferID}`
+      if (verbose) {
+        url = url + '?verbose=true'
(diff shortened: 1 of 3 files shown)


keiko713:
@ransombriggs I think following `get` doesn't expect `logs` to be `[]` because you're not passing `verbose`:

https://github.com/heroku/heroku-pg/blob/610112d1044d55b95335c88a9afd33afef3af9a5/test/commands/backups/capture.js#L29

ransombriggs:
@keiko713 thanks, pushed up https://github.com/heroku/heroku-pg/pull/76/commits/42afccf7d0bef790ee5e3a223830acf6b4b62686 to reflect the logs omission without verbose

keiko713:
lgtm, ship it! Seems like we have better test with restore than capture, but we can just improve this later.

keiko713 (approved):

PR #80, opened 2016-12-16
Show load properly for pg:diagnose

keiko713:
Seems like showing the certain value (only `Load`, I believe) that you get from pgdiagnose is not working properly. Here is expected response json examples (it's in the test as well):

```
  {
    "name": "Hit Rate",
    "status": "red",
    "results": [
      {
        "name": "overall index hit rate",
        "ratio": 0.9861242393582039
      },
      {
        "name": "overall cache hit rate",
        "ratio": 0.9637371701579502
      }
    ]
  },
  {
    "name": "Blocking Queries",
    "status": "green",
    "results": null
  },
  {
    "name": "Sequences",
    "status": "green",
    "results": null
  },
  {
    "name": "Load",
    "status": "red",
    "results": {
      "load": "15.36"
    }
  }
```

Previously, in Ruby version, [it was taken care of this way](https://github.com/heroku/legacy-cli/blob/d7e9718079b9936ffd264529bdd03a6734fd64cc/lib/heroku/helpers/pg_diagnose.rb#L92-L100), which wasn't converted to Node properly.

Ideally, it'll be like;

```
RED: Load
Load 100
```

instead of;

```
RED: Load
load 100
```

but seems like Node doesn't have any easy way to capitalize the first letter so...

Can you please review @ransombriggs ? Thanks!

### DIFF
--- a/commands/diagnose.js
+++ b/commands/diagnose.js
@@ -2,6 +2,7 @@
 
 const cli = require('heroku-cli-util')
 const co = require('co')
+const capitalize = require('lodash.capitalize')
 const PGDIAGNOSE_HOST = process.env.PGDIAGNOSE_URL || 'https://pgdiagnose.herokai.com'
 
 function * run (context, heroku) {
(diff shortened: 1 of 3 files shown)


keiko713:
@ransombriggs your part lgtm ! and nice to have more tests!

ransombriggs (approved):

PR #81, opened 2017-01-05
Add relevant documentation for pg:pull

slant:
The examples in the documentation for `pg:pull` were examples for `pg:push`.

I should be clear that I am assuming that the explicit URI in the second example is valid. Please test this before merging this, until you already know that it works.

### DIFF
--- a/commands/pull.js
+++ b/commands/pull.js
@@ -188,10 +188,10 @@ TARGET will be created locally if it's a database name or remotely if it's a ful
 Examples:
 
   # pull Heroku DB named postgresql-swimmingly-100 into local DB mylocaldb
-  $ heroku pg:push postgresql-swimmingly-100 mylocaldb
+  $ heroku pg:pull HEROKU_POSTGRESQL_MAGENTA mylocaldb --app sushi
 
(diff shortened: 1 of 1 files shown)


commands/pull.js
```diff
@@ -188,10 +188,10 @@ TARGET will be created locally if it's a database name or remotely if it's a ful
 Examples:
 
   # pull Heroku DB named postgresql-swimmingly-100 into local DB mylocaldb
-  $ heroku pg:push postgresql-swimmingly-100 mylocaldb
+  $ heroku pg:pull HEROKU_POSTGRESQL_MAGENTA mylocaldb --app sushi
```
keiko713: I *think* we are trying to get people more familiar with addon haiku name than color, so maaaybe the example can stay using haiku name. I'm not super positive about it, though.
Also, if we're going to change here, we should change the example of `pg:push` as well.

ransombriggs: @keiko713 changed color back to haiku in 67d9ae6

ransombriggs:
@slant I verified the example commands with my local postgres, will merge & release

ransombriggs (approved):

PR #86, opened 2017-01-24
Request the backup with logs on failure

ransombriggs:


mathias:
If we're using a ternary here, could it just be like this?

```javascript
let url = quietUrl + verbose ? '?verbose=true' : ''
```

I'm not sure if that is really that readable though?

ransombriggs:
@mathias I need the `verboseUrl` down in the error handling, which is why I structured the ternary this way, so I would not have to repeat the part that makes it verbose.

mathias:
@ransombriggs Sounds good, makes sense.

mathias (approved):

PR #89, opened 2017-02-07
added missing help descriptions

jdx:


commands/wait.js
```diff
@@ -57,6 +57,7 @@ function * run (context, heroku) {
 module.exports = {
   topic: 'pg',
   command: 'wait',
+  description: 'blocks until database is created',
```
msakrejda: "is available" maybe? Technically they're created right away.

jdx: sure, the old message was `monitor database creation, exit when complete`, which I didn't think was very good at all

msakrejda:
+1

Issue #85, opened 2017-01-24
allow connection to local DB through Unix socket


derikson:
I don't see that there is currently any way to connect to a database through a Unix socket with `heroku pg:pull`.

The PostgreSQL command line tools connect using a socket if the host part is empty or starts with a slash. But this command changes the host to `localhost` if it's empty, and the host portion of a URI can only contain a slash if it's percent-encoded, but `url.parse()` treats that as part of the path name.

See https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING for reference.

msakrejda:
I believe this was fixed by #97; please let me know if you are still having problems.

PR #97, opened 2017-02-23
Fix env cascading when invoking psql

### LINKED ISSUE #92: False extension mismatch reported with pg:pull + partially-qualified local database + no ssl
I have a local database, `sample`, that I am `pg:pull`-ing into. The local database server doesn't support SSL connections, and the production database has 7 extensions.

If I use `heroku pg:pull DATABASE sample`, the server's extensions will be correctly `CREATE`-ed locally, but the CLI will report an extension mismatch:

```
heroku-cli: Pulling postgresql-removed-999999 ---> sample
pg_dump: reading extensions
pg_dump: identifying extension members
pg_dump: reading schemas
[...]
pg_restore: creating SCHEMA "public"
pg_restore: creating EXTENSION "plpgsql"
pg_restore: creating COMMENT "EXTENSION plpgsql"
pg_restore: creating EXTENSION "fuzzystrmatch"
pg_restore: creating COMMENT "EXTENSION fuzzystrmatch"
pg_restore: creating EXTENSION "hstore"
pg_restore: creating COMMENT "EXTENSION hstore"
pg_restore: creating EXTENSION "pg_stat_statements"
pg_restore: creating COMMENT "EXTENSION pg_stat_statements"
pg_restore: creating EXTENSION "pg_trgm"
pg_restore: creating COMMENT "EXTENSION pg_trgm"
pg_restore: creating EXTENSION "unaccent"
pg_restore: creating COMMENT "EXTENSION unaccent"
pg_restore: creating EXTENSION "uuid-ossp"
pg_restore: creating COMMENT "EXTENSION "uuid-ossp""
[...]
psql: server does not support SSL, but SSL was required
 ▸    WARNING: Extensions in newly created target database differ from existing source database.
 ▸    Target extensions:
 ▸    
 ▸    Source extensions:
 ▸    extname
 ▸    --------------------
 ▸    fuzzystrmatch
 ▸    hstore
 ▸    pg_stat_statements
 ▸    pg_trgm
 ▸    plpgsql
 ▸    unaccent
 ▸    uuid-ossp
 ▸    (7 rows)
 ▸    
 ▸    
 ▸    HINT: You should review output to ensure that any errors
 ▸    ignored are acceptable - entire tables may have been missed, where a dependency
 ▸    could not be resolved. You may need to to install a postgresql-contrib package
 ▸    and retry.
heroku-cli: Pulling complete.
```

If I fully-qualify my local database's URL with `heroku pg:pull DATABASE postgres://localhost/sample`, however, it works:

```
heroku-cli: Pulling postgresql-removed-999999 ---> sample
pg_dump: reading extensions
pg_dump: identifying extension members
pg_dump: reading schemas
[...]
pg_restore: creating SCHEMA "public"
pg_restore: creating EXTENSION "plpgsql"
pg_restore: creating COMMENT "EXTENSION plpgsql"
pg_restore: creating EXTENSION "fuzzystrmatch"
pg_restore: creating COMMENT "EXTENSION fuzzystrmatch"
pg_restore: creating EXTENSION "hstore"
pg_restore: creating COMMENT "EXTENSION hstore"
pg_restore: creating EXTENSION "pg_stat_statements"
pg_restore: creating COMMENT "EXTENSION pg_stat_statements"
pg_restore: creating EXTENSION "pg_trgm"
pg_restore: creating COMMENT "EXTENSION pg_trgm"
pg_restore: creating EXTENSION "unaccent"
pg_restore: creating COMMENT "EXTENSION unaccent"
pg_restore: creating EXTENSION "uuid-ossp"
pg_restore: creating COMMENT "EXTENSION "uuid-ossp""
[...]
heroku-cli: Pulling complete.
```

Note the line `psql: server does not support SSL, but SSL was required` in the failing example.

The psql code here is not correctly checking for `null` hostnames, which imply `localhost` in Postgres URLs: https://github.com/heroku/heroku-pg/blob/e9e17e64b8b9a819127259d6482c542245ee3dac/lib/bastion.js#L29

URLs like `foo` are valid Postgres URLs: `foo` is equivalent to `postgres://localhost/foo` (default user, no password, default port, etc)

msakrejda:
Currently we override most PG* environment variables, but there's no
good reason for this--if the user has set something, we should not
require them to override this via the URL.

With this change, they still *can* if they need to do so explicitly,
but we shouldn't clobber valid various from the environment with empty
ones.

We keep PGSSLMODE as a setting users cannot override (to avoid
unencrypted connections to remote databases), but account for unset
hosts (which defaults to a local connection).

Fixes #92

### DIFF
--- a/lib/bastion.js
+++ b/lib/bastion.js
@@ -24,15 +24,25 @@ const getBastion = function (config, baseName) {
 exports.getBastion = getBastion
 
 const env = function (db) {
-  return Object.assign({}, process.env, {
-    PGAPPNAME: 'psql non-interactive',
-    PGSSLMODE: db.hostname === 'localhost' ? 'prefer' : 'require',
-    PGUSER: db.user || '',
-    PGPASSWORD: db.password,
-    PGDATABASE: db.database,
-    PGPORT: db.port || 5432,
-    PGHOST: db.host
+  let baseEnv = Object.assign({
+    PGAPPNAME: 'psql non-interactive'
+  }, process.env, {
+    PGSSLMODE: (!db.hostname || db.hostname === 'localhost') ? 'prefer' : 'require'
   })
+  let mapping = {
+    PGUSER: 'user',
+    PGPASSWORD: 'password',
+    PGDATABASE: 'database',
+    PGPORT: 'port',
+    PGHOST: 'host'
+  }
+  Object.keys(mapping).forEach((envVar) => {
+    let val = db[mapping[envVar]]
+    if (val) {
+      baseEnv[envVar] = val
+    }
+  })
+  return baseEnv
 }
 
 exports.env = env


msakrejda:
I think this fixes #85 as well (you'll be able to either set  a `PGHOST` env var or leave the host blank and your default host is probably Unix socket if your OS supports that).

fnando:
Looks legit! I didn't test it though.

+1

fnando (approved):

michaelbaudino:
@uhoh-itsmaciek I see at least one use-case where one would need to override `PGSSLMODE`: when using the official Postgres Docker image for development (which does not support SSL).

The hostname is then considered remote (it's usually a different container than the one running the app, thus usually in Docker virtual network), but cannot be access neither via Unix socket nor with SSL :cry: 

I'd be happy to provide a PR that makes `PGSSLMODE` overridable like any other `PG*` variables, but I'd like to know if it might be accepted before investing time in it.

Unless you have another workaround I didn't think about?

msakrejda:
@michaelbaudino yeah, okay, I think that's safe enough. Heroku Postgres will refuse unencrypted connections anyway, so it's not much of an issue. You plan to just move PGSSLMODE [here](https://github.com/heroku/heroku-pg/pull/97/files#diff-591dc38b08ab723474d07f3fe17e172fR28), correct? That works for me...

michaelbaudino:
Woohoo, thanks, here is the PR: #171 :heart_eyes:

PR #98, opened 2017-03-25
Address the edge case of async provisioning

keiko713:
Right now, after the Postgres started provisioning, `heroku pg` command will return following over time:

**0. Provision**

```
$ heroku addons:create heroku-postgresql:standard-0 -a keiko-heroku-pg-test
Creating heroku-postgresql:standard-0 on ⬢ keiko-heroku-pg-test... $50/month
The database should be available in 3-5 minutes.
! CAUTION: The database will be empty. If upgrading, you can transfer
!          data from another database with pg:copy.
Use `heroku pg:wait` to track status.
postgresql-acute-75514 is being created in the background. The app will restart when complete...
Use heroku addons:info postgresql-acute-75514 to check creation progress
Use heroku addons:docs heroku-postgresql to view documentation
```

**1. Right after the provision and shogun is still returning 404**

```
$ heroku pg -a keiko-heroku-pg-test
 ▸    postgresql-acute-75514 is not yet provisioned.
 ▸    Run heroku pg:wait to wait until the db is provisioned.
```

**2. Provisioning is ongoing and shogun is not returning 404 anymore**

```
$ heroku pg -a keiko-heroku-pg-test
=== 
Plan:               Standard 0
Status:             Preparing
Fork/Follow:        Temporarily Unavailable
Rollback:           Temporarily Unavailable
Created:            2017-03-25 00:33 UTC
Region:             us
Maintenance:        not required
Maintenance window: Mondays 19:00 to 23:00 UTC
Add-on:             postgresql-acute-75514
```

**3. Provision has finished**

```
$ heroku pg -a keiko-heroku-pg-test
=== DATABASE_URL
Plan:               Standard 0
Status:             Available
Fork/Follow:        Available
Rollback:           earliest from 2017-03-25 00:35 UTC
Created:            2017-03-25 00:33 UTC
Region:             us
Maintenance:        not required
Maintenance window: Mondays 19:00 to 23:00 UTC
Add-on:             postgresql-acute-75514
```

This is confusing as the users will see the result with empty header with step 2. With this PR, instead of using config var approach and use the attachment names to show the header.

NOTE: I tried to implement this with `heroku pg:info some-specific-db` but it gets too complicated and kinda gave up. I'd like to hear if there is any way to improve this part as well.

### DIFF
--- a/commands/info.js
+++ b/commands/info.js
@@ -5,9 +5,14 @@ const cli = require('heroku-cli-util')
 const util = require('../lib/util')
 
 function displayDB (db, app) {
-  cli.styledHeader(db.configVars.map(c => cli.color.configVar(c)).join(', '))
+  if (db.addon.attachment_names) {
+    cli.styledHeader(db.addon.attachment_names.map(c => cli.color.configVar(c + '_URL')).join(', '))
+  } else {
+    cli.styledHeader(db.configVars.map(c => cli.color.configVar(c)).join(', '))
+  }
+
   if (db.addon.app.name !== app) {
-    db.db.info.push({name: 'Billing App', values: [db.addon.app.name]})
+    db.db.info.push({name: 'Billing App', values: [cli.color.cyan(db.addon.app.name)]})
   }
   db.db.info.push({name: 'Add-on', values: [cli.color.addon(db.addon.name)]})
   let info = db.db.info.reduce((info, i) => {
@@ -56,7 +61,7 @@ function * run (context, heroku) {
         path: `/client/v11/databases/${addon.name}`
       }).catch(err => {
         if (err.statusCode !== 404) throw err
-        cli.warn(`${cli.color.addon(addon.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku pg:wait')} to wait until the db is provisioned.`)
+        cli.warn(`${cli.color.addon(addon.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
       })
     }
   })
--- a/lib/fetcher.js
+++ b/lib/fetcher.js
@@ -66,6 +66,7 @@ module.exports = heroku => {
       }
     }
 
+    // case for multiple attachments with passedDb
     let first = matches[0]
 
     // case for 422 where there are ambiguous attachments that are equivalent
@@ -101,13 +102,32 @@ module.exports = heroku => {
     return attachments.filter(a => a.addon.plan.name.startsWith('heroku-postgresql'))
   }
 
+  function * getAttachmentNamesByAddons (attachments) {
+    let results = {}
+    attachments.forEach(function (a) {
+      let attachmentNames = results[a.addon.id] || []
+      attachmentNames = attachmentNames.concat(a.name)
+      results[a.addon.id] = attachmentNames
+    })
+
+    return results
+  }
+
   function * all (app) {
     const uniqby = require('lodash.uniqby')
 
     debug(`fetching all DBs on ${app}`)
 
-    let addons = (yield allAttachments(app)).map(a => a.addon)
+    let attachments = yield allAttachments(app)
+    let addons = attachments.map(a => a.addon)
+
+    // Get the list of config vars per addon here and add to each addon
+    let attachmentNamesByAddons = yield getAttachmentNamesByAddons(attachments)
     addons = uniqby(addons, 'id')
+    addons = addons.map(function (a) {
+      a.attachment_names = attachmentNamesByAddons[a.id]
+      return a
+    })
 
     return addons
   }
--- a/lib/fetcher.js
+++ b/lib/fetcher.js
@@ -102,7 +102,7 @@ module.exports = heroku => {
     return attachments.filter(a => a.addon.plan.name.startsWith('heroku-postgresql'))
   }
 
-  function * getAttachmentNamesByAddon (attachments) {
+  function getAttachmentNamesByAddon (attachments) {
     return attachments.reduce((results, a) => {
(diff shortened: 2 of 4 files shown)


keiko713:
Ideally, I would like to stop using referring config var for any sort of `heroku pg` commands if possible.
For example, if you just copy and paste the database connection info config var and create a new one, it'll show up in `heroku pg` currently. This is confusing because Heroku (or Postgres) doesn't know about it as "attachment" and Heroku Postgres won't update such config var for any type of maintenance, still yet it gives customers the wrong impression that it's "Heroku certified" config vars as it'll show up in `heroku pg`.
It'd be even nicer if we can warn such things in CLI, with the correct attachment command to fix the thing.

keiko713:
Forgot to write some background. This (step 2) had been working okay, but due to the provisioning change that we made this week (async provisioning with Heroku Postgres), it is becoming "an empty" header. With async provisioning, the application is not going to get new config vars during the provisioning. It'll be only available once it's "provisioned", which would take a few minutes.

commands/info.js
```diff
@@ -56,7 +61,7 @@ function * run (context, heroku) {
         path: `/client/v11/databases/${addon.name}`
       }).catch(err => {
         if (err.statusCode !== 404) throw err
-        cli.warn(`${cli.color.addon(addon.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku pg:wait')} to wait until the db is provisioned.`)
+        cli.warn(`${cli.color.addon(addon.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
```
msakrejda, line 64: Is `pg:wait` obsolete with `addons:wait`? Should we flag it as deprecated and redirect users?

keiko713: Good question. I realized that `pg:wait` doesn't work well under certain circumstances, so especially when people want to know the status of provisioning, `addons:wait` is the safest option at the moment. However, `pg:wait` doesn't do exactly the same thing as `addons:wait`. For example, `pg:wait` tells you more detailed progress when you created an Postgres addon as fork or follow (https://github.com/heroku/shogun/blob/master/lib/shogun/presenters/resource_client_api_presenter.rb#L125-L149). I'm also thinking to tweak [this provisioning message](https://github.com/heroku/shogun/blob/master/lib/web/mediators/formations/postgres_provisioner.rb#L372) and use `heroku addons:wait`, but I'm not sure if we should "deprecate" `pg:wait` for that reason.

keiko713: I also meant to look into more for exactly what situation `pg:wait` doesn't work well and potentially fix it, but I also have some feeling that it may not be an easy fix.

lib/fetcher.js
```diff
@@ -101,13 +102,32 @@ module.exports = heroku => {
...
 
+  function * getAttachmentNamesByAddons (attachments) {
+    let results = {}
+    attachments.forEach(function (a) {
+      let attachmentNames = results[a.addon.id] || []
+      attachmentNames = attachmentNames.concat(a.name)
+      results[a.addon.id] = attachmentNames
+    })
+
+    return results
```
msakrejda: This is fine, but you could simplify this to

```javascript
return attachments.reduce((results, a) => {
  results[a.addon.id] = (results[a.addon.id] || []).concat(a.name)
  return results
}, {})
```

or even jus tdrop line 109 and move the `.concat(a.name)` to 110.

lib/fetcher.js
```diff
@@ -101,13 +102,32 @@ module.exports = heroku => {
     return attachments.filter(a => a.addon.plan.name.startsWith('heroku-postgresql'))
   }
 
+  function * getAttachmentNamesByAddons (attachments) {
```
msakrejda: I think this should be `Addon` singular: `getAttachmentNamesByAddon`.

lib/fetcher.js
```diff
@@ -101,13 +102,32 @@ module.exports = heroku => {
...
     debug(`fetching all DBs on ${app}`)
 
-    let addons = (yield allAttachments(app)).map(a => a.addon)
+    let attachments = yield allAttachments(app)
+    let addons = attachments.map(a => a.addon)
+
+    // Get the list of config vars per addon here and add to each addon
+    let attachmentNamesByAddons = yield getAttachmentNamesByAddons(attachments)
     addons = uniqby(addons, 'id')
+    addons = addons.map(function (a) {
```
msakrejda: I'd chain this call with the `uniqby` above--`addons = uniqby(addons, 'id').map...`.

msakrejda: And actually since you're mutating the objects anyway, maybe make this a `forEach` instead of a map to make that clear?

msakrejda (approved):
Seems legit. +1 as long as it works in staging.

keiko713:
I tested with `heroku plugins:link` and here is result:

Before (without `heroku plugins:link`):

```
$ h pg -r test
=== 
Plan:               Private 0
Status:             Preparing
HA Status:          Temporarily Unavailable
Fork/Follow:        Temporarily Unavailable
Rollback:           Temporarily Unavailable
Created:            2017-03-29 18:49 UTC
Region:             virginia
Maintenance:        not required
Maintenance window: Wednesdays 21:30 to Thursdays 01:30 UTC
Add-on:             postgresql-keiko-reticulated-34216
Credentials:        2
```

After:

```
$ h pg -r test
=== HEROKU_POSTGRESQL_KEIKO_URL
Plan:               Private 0
Status:             Preparing
HA Status:          Temporarily Unavailable
Fork/Follow:        Temporarily Unavailable
Rollback:           Temporarily Unavailable
Created:            2017-03-29 18:49 UTC
Region:             virginia
Maintenance:        not required
Maintenance window: Wednesdays 21:30 to Thursdays 01:30 UTC
Add-on:             postgresql-keiko-reticulated-34216
```

PR #99, opened 2017-03-28
Fail psql if the tunnel fails

msakrejda:
This is especially important since the tunnel promise can resolve even
if the authentication information is incorrect, which leads to an ugly
stack trace propagated to the client.


camillebaldock (approved):
thanks for this

PR #103, opened 2017-03-31
Move all commands to use add-on uuids instead of names

msakrejda:
This is more consistent with DHC and the Kafka CLI and works around
issues around add-on renames and Shogun sync.


msakrejda:
I've verified the changes still work.

fnando:
Looks solid! This is good to have! \m/

+1

fnando (approved):

PR #106, opened 2017-04-07
Don't force hostname of localhost

tpope:
Using `-h localhost` prevents the use of a socket file, when in turn requires the user to enable network connections and set up password authentication.


cbartlett (approved):

msakrejda (approved):
Looks great, thanks!

mathias:
Thank you @tpope !

PR #107, opened 2017-04-13
Psql drop timeout

### LINKED ISSUE #94: heroku pg:psql -c 'select pg_sleep(100)' will fail in the middle due to 20s timeout
https://github.com/heroku/heroku-pg/blob/master/lib/psql.js#L19

Also, seems like you can't really change this timeout, and I'm not really sure where 20s is coming from. Well, maybe for bastion? but still...

msakrejda:
This causes problems for long-running queries, and is only really
useful for a slightly nicer experience in the rare cases when the
bastion or service is unavailable.

Fixes #94.

The second commit fixes tests. Some of these had bad mocking and seemed to rely on the timeout to pass. This fixes them and removes the timeout catch.

### DIFF
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -13,24 +13,22 @@ function handlePsqlError (reject, psql) {
   })
 }
 
-function execPsql (query, dbEnv, timeout) {
+function execPsql (query, dbEnv) {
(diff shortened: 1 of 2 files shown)


idan (approved):
Nice. +1

PR #108, opened 2017-05-11
trim package size

jdx:


msakrejda:
Nice, thanks! Do you know why coverage appears to have dropped? I can't make sense of it based on the diff...

jdx:
yeah it shouldn't have impacted it, I just tried to rebuild the project to see if it might change. Best guess is mocha is randomizing the test order and that took it down a slightly different code path?

msakrejda (approved):
Huh, okay. Cleraly not an issue with the patch, so +1

PR #109, opened 2017-05-16
Credentials commands

camillebaldock:
Postgres credentials will be released soon as a private beta.
This adds support for the pg side of the commands for that project.
The commands will return `Authorised users only - credentials private beta.` for non-whitelisted addon owners. `pg:credentials` is backwards compatible with previous behaviour (it exhibits the new behaviour for whitelisted users, old behaviour for non-whitelisted users)


camillebaldock:
Added more tests to `credentials.js` to go over both behaviours and correct the loss in coverage reported above ^^

test/commands/info.js
```diff
@@ -83,13 +83,11 @@ Plan:        Hobby-dev
 Following:   HEROKU_POSTGRESQL_COBALT
 Billing App: myapp2
 Add-on:      postgres-1
-Credentials: 2
```
msakrejda: So the API is conditionally returning this for flagged-in users now, right?

camillebaldock: yes that's right

commands/credentials.js
```diff
@@ -11,22 +11,23 @@ function * run (context, heroku) {
...
+                                       { host: host(addon) })
       cli.table(credentials, {
         columns: [
           {key: 'name', label: 'Credential'},
           {key: 'state', label: 'State'}
         ]
       })
-    else
+    } catch (err) {
+      if (err.statusCode !== 422) throw err
```
msakrejda: 422 feels weird here (since this is a GET and there is no body), but given this is temporary anyway, it's not worth nitpicking.

camillebaldock: yes we should go for a generic 400 here now that I've extended the error to cover GET endpoints too: happy to change

commands/credentials.js
```diff
@@ -9,11 +9,25 @@ function * run (context, heroku) {
...
+    try {
+      let credentials = yield heroku.get(`/postgres/v0/databases/${addon.name}/credentials`,
+                                       { host: host(addon) })
+      cli.table(credentials, {
+        columns: [
+          {key: 'name', label: 'Credential'},
+          {key: 'state', label: 'State'}
+        ]
+      })
+    } catch (err) {
```
msakrejda, line 23: So we can't do much about not supporting `--reset` here for the new interface, right? We can drop it once the feature is public?

commands/credentials/repair_default.js
```diff
@@ -0,0 +1,33 @@
...
+
+function * run (context, heroku) {
+  const fetcher = require('../../lib/fetcher')(heroku)
+  const host = require('../../lib/host')
+  const util = require('../../lib/util')
+
+  const {app, args} = context
+
+  let db = yield fetcher.addon(app, args.database)
+  if (util.starterPlan(db)) throw new Error('This operation is not supported by Hobby tier databases.')
```
msakrejda, line 14: I wonder if we should explain why? I.e., that they don't have configurable permissions in the first place... it's fine for now, though; we can address that during the beta if there's confusion...

genslein:
General knowledge question, can we get codecov for our other projects? 😏

commands/credentials.js
```diff
@@ -9,11 +9,25 @@ function * run (context, heroku) {
...
+      if (err.statusCode !== 422) throw err
```
jdx: err might not be an http error, you should throw in that case probably

msakrejda: Technically `err.statusCode` is `undefined` in that case, which `!== 422`, right? =D

But yeah, we may want to treat that case more explicitly...

commands/credentials/destroy.js
```diff
@@ -0,0 +1,47 @@
...
+
+  const {app, args, flags} = context
+  let cred = flags.name
+
+  let db = yield fetcher.addon(app, args.database)
+  if (util.starterPlan(db)) throw new Error('This operation is not supported by Hobby tier databases.')
+
+  let attachments = yield heroku.get(`/addons/${db.name}/addon-attachments`)
+  let credAttachments = attachments.filter(a => a.namespace === `credential:${flags.name}`)
+  if (credAttachments.length > 0) throw new Error(`Credential ${flags.name} must be detached from all other apps before destroying.`)
```
msakrejda, line 19: Not blocking, but we could probably improve this message by listing the apps or attachments so it's clear what to detach.

commands/credentials/url.js
```diff
@@ -0,0 +1,59 @@
...
+function * run (context, heroku) {
+  const host = require('../../lib/host')
+  const fetcher = require('../../lib/fetcher')(heroku)
+  const util = require('../../lib/util')
+
+  const {app, args, flags} = context
+
+  let db = yield fetcher.addon(app, args.database)
+  let cred = flags.name || 'default'
+  if (util.starterPlan(db) && cred !== 'default') throw new Error('This operation is not supported by Hobby tier databases.')
```
msakrejda, line 16: This error message is slightly misleading: the action is supported; just not with that option.

msakrejda (approved):
Some minor comments, but looks great. +1

commands/credentials.js
camillebaldock: that's right, we will be deprecating it in favour of `pg:credentials:rotate --all` in the future; there are plans to add a deprecation warning, communicate with customers who we know are using it in automated scripts: this is still early days though

commands/credentials/destroy.js
camillebaldock: There actually is already a story for that :) working through a few higher priority items at the moment, but expect a patch in a few days :)

commands/credentials.js
jdx: yep, you're right @uhoh-itsmaciek. This is fine, but to avoid confusion (like I just had) might be better to write `if (!err.statusCode || err.statusCode !== 422)`

commands/credentials/url.js
camillebaldock: there's another card to deal with confusing hobby message credentials in which I will fix that :). you're right we can do better here

commands/credentials/repair_default.js
camillebaldock: yes we should, i'm taking this on in the card to fix hobby-dev calls messaging

commands/credentials.js
camillebaldock: fixed :)

PR #111, opened 2017-05-17
Add attachment info to pg:credentials

camillebaldock:
## What is this?

* Provide attachment information in the list of credentials
* Order default first

## Notes

This PR is based on the `pg-credentials` branch, once the `pg-credentials` branch is merged, I will point it to master.

### DIFF
--- a/commands/credentials.js
+++ b/commands/credentials.js
@@ -2,6 +2,8 @@
 
 const co = require('co')
 const cli = require('heroku-cli-util')
+const sortBy = require('lodash.sortby')
+const printf = require('printf')
 
 function * run (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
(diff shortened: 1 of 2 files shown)


genslein (approved):

commands/credentials.js
```diff
@@ -11,12 +13,57 @@ function * run (context, heroku) {
...
+      )
+
+      // render each attachment under the credential
+      let attLines = atts.map(function (attachment, idx) {
+        let isFirst = (idx === credAttachments.length - 1)
+        return renderAttachment(attachment, app, isFirst)
+      })
+
+      return [cred].concat(attLines).join('\n')
+    }
```
msakrejda: This seems pretty fiddly. The code used by `addons` to do basically the same thing is not reusable here?

commands/credentials.js
```diff
@@ -11,12 +13,57 @@ function * run (context, heroku) {
...
+      let isForeignApp = (attOrAddon) => attOrAddon.app.name !== app
+      let atts = sortBy(credAttachments,
+        isForeignApp,
+        'name',
+        'app.name'
+      )
+
+      // render each attachment under the credential
+      let attLines = atts.map(function (attachment, idx) {
+        let isFirst = (idx === credAttachments.length - 1)
```
msakrejda: Technically `isLast`, no?

msakrejda (approved):
+1

commands/credentials.js
camillebaldock: renderAttachment is straight copy from there, unfortunately the ordering of credentials if differerent that in addons, and has the special case of `default` so it wasn't quite 100% reusable; where would be put shared-ish code between those CLI plugins?

commands/credentials.js
camillebaldock: yes!, funnily enough that wrong name is straight from heroku-cli-addons: fixing it here

PR #113, opened 2017-05-18
2.1.0 release

camillebaldock:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "heroku-pg",
   "description": "Heroku CLI plugin to manage Postgres.",
-  "version": "2.0.26",
+  "version": "2.0.27",
```
msakrejda: This is for the creds changes, yeah? Shouldn't that be a `2.1.0`?

msakrejda (approved):
Minor suggestion, but +1

package.json
camillebaldock: it doesn't break any of the existing behaviour but it does add a lot of new commands so yes agreed, i'm updating now

camillebaldock:
@dickeyxxx hi! what is the process for releasing updated npm packages for a new heroku-pg version?

camillebaldock:
Has been released

PR #115, opened 2017-05-18
Repeat gross hack from 67d874ed0b4c977bbc96e5d9289f48571443dc58

### LINKED ISSUE #114: pg:ps fails with 'column "waiting" does not exist'
i'm running this command against heroku postgres 9.6.x databases. i'm running from OSX, heroku CLI installed via brew. 
```
heroku pg:ps --app rivaliq-usertest
ERROR:  column "waiting" does not exist
LINE 7:  waiting,
         ^
Timing is on.
Expanded display is used automatically.
Null display is "(null)".
Time: 73.484 ms
```

```
heroku --version
heroku-cli/6.6.13-57fc242 (darwin-x64) node-v7.9.0

which heroku
/usr/local/bin/heroku

ls -l /usr/local/bin/heroku
lrwxr-xr-x  1 spollack  admin  33 May 18 13:17 /usr/local/bin/heroku -> ../Cellar/heroku/6.6.7/bin/heroku

cat /usr/local/Cellar/heroku/6.6.7/libexec/node_modules/heroku-pg/package.json | grep "version"
  "version": "2.0.26",
```

I see there was a fix here in heroku-pg to be compatible with pg 9.6.x, but it doesn't seem to be working for me. 

Thanks,
Seth

msakrejda:
This is pretty awful but it's reasonably robust. A better fix probably involves introducing a real Postgres driver rather than trying to coral psql into acting like one for us.

Fixes #114

### DIFF
--- a/commands/ps.js
+++ b/commands/ps.js
@@ -13,20 +13,20 @@ function * run (context, heroku) {
 
   let db = yield fetcher.database(app, database)
 
+  const num = Math.random()
+  const waitingMarker = `${num}${num}`
+
   let waitingQuery = `
-SELECT EXISTS(
+SELECT '${num}' || '${num}' WHERE EXISTS (
   SELECT 1 FROM information_schema.columns WHERE table_schema = 'pg_catalog'
     AND table_name = 'pg_stat_activity'
     AND column_name = 'waiting'
-) AS available
+)
 `
-
   let waitingOutput = yield psql.exec(db, waitingQuery)
-  let waiting = 'wait_event IS NOT NULL AS waiting'
-
-  if (waitingOutput.includes('t')) {
-    waiting = 'waiting'
-  }
+  let waiting = waitingOutput.includes(waitingMarker)
+                ? 'waiting'
+                : 'wait_event IS NOT NULL AS waiting'
 
   let query = `
 SELECT


fnando (approved):
I don't know what's happening here, so yay?

msakrejda:
@fnando basically the old code does something like `execSync('psql -c "select ..."')` (where the select returns a boolean) and then looks for `t` in the result (psql's default rendering of `true`). An expected response on stdout look something like this:

```console
maciek@mothra:~$ psql --no-psqlrc -c "select true as available" 2>/dev/null
 available 
-----------
 t
(1 row)
```

So yes, if everything goes as expected, that does contain `t`, `includes('t')` is true, and everything is fine. 

However, there are a hundred different reasons from localization to various `.psqlrc` settings to FSM knows what for a `t` to be returned in the body of the response aside from the query evaluating to `true`, so this is a very bad way to do this. However, fixing this properly involves bringing in the `pg` module and retrofitting a bunch of call sites, so as a more robust hack, I'm constructing a string that's vanishingly unlikely to occur in any actual response, then concatenating it to itself (so that it's never echoed in an error that includes the text of the query), and only producing it in the response if the condition holds.

PR #119, opened 2017-05-23
Better error message when destroying a credential with attachments

camillebaldock:
Before:
```
heroku pg:credentials:destroy --name hello -a camille-main-app

 ▸    Credential hello must be detached from all other apps before destroying.
```

After:
```
heroku pg:credentials:destroy --name hello -a camille-main-app

 ▸    Credential hello must be detached from the apps ⬢ camille-secondary-app, ⬢ camille-main-app before destroying.
```


msakrejda (approved):
Excellent. +1

PR #120, opened 2017-05-26
Add np for more convenient releases

camillebaldock:


package.json
```diff
@@ -29,6 +29,7 @@
     "heroku-client": "3.0.2",
     "mocha": "3.2.0",
     "nock": "9.0.11",
+    "np": "^2.15.0",
```
msakrejda: We should save the exact version here, no? Also @dickeyxxx introduced yarn.lock in e8dc71e9749e138710cec3afe016397dfa3cfeb3 -- we should probably use yarn to add this dep and updated the lockfile.

msakrejda (changes requested):

package.json
jdx: typically we save exact versions, so yeah, it's a good idea. In practice though, you could either edit the yarn.lock file OR make this exact and it would achieve the same result (though not for transitive dependencies)

camillebaldock:
@uhoh-itsmaciek pegged the version in package.json and used yarn for dependencies this time round, would love to 🍐 if this isn't what you expected so I can understand what the dependencies workflow should be on these CLI projects

camillebaldock:
related: https://github.com/heroku/heroku-cli-addons/pull/82

msakrejda (approved):
Looks great, +1

PR #121, opened 2017-05-26
Should not combine --force and --all

camillebaldock:
Before:

`heroku pg:credentials:rotate --force --all -a camille-hobby-app`

After:

`Cannot force rotate the default credential.`

We should not allow `--force` with `--all` since we do not support force rotation of the default credential.

### DIFF
--- a/commands/credentials/rotate.js
+++ b/commands/credentials/rotate.js
@@ -16,7 +16,11 @@ function * run (context, heroku) {
   }
   let cred = flags.name || 'default'
   if ((cred === 'default' || all) && 'force' in flags) {
-    throw new Error('Cannot force rotate the default credential.')
+    if (all) {
+      throw new Error('Cannot force rotate all credentials: the default credential cannot be force rotated.')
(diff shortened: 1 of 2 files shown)


commands/credentials/rotate.js
```diff
@@ -12,11 +12,11 @@ function * run (context, heroku) {
...
 
   if (all && 'name' in flags) {
-    cli.exit(1, 'cannot pass both --all and --name')
+    throw new Error('cannot pass both --all and --name')
   }
   let cred = flags.name || 'default'
-  if (cred === 'default' && 'force' in flags && !all) {
-    cli.exit(1, 'Cannot force rotate the default credential.')
+  if ((cred === 'default' || all) && 'force' in flags) {
+    throw new Error('Cannot force rotate the default credential.')
```
msakrejda: The message doesn't quite match the `all` case but close enough I guess.

msakrejda (approved):
+1

PR #122, opened 2017-05-30
Include username and transaction start in pg:ps output

camillebaldock:
Before:

```
heroku pg:ps -a camille-main-app
 pid | state  |        source        |   running_for   | waiting |        query         
-----+--------+----------------------+-----------------+---------+----------------------
 198 | active | psql non-interactive | 00:00:06.198973 | f       | SELECT pg_sleep(10);
```

After:

```
heroku pg:ps -a camille-main-app
 pid | state  |        source        |    usename     |   running_for   | waiting |        query         
-----+--------+----------------------+----------------+-----------------+---------+----------------------
 198 | active | psql non-interactive | u8l0buj5sjbsj4 | 00:00:06.198973 | f       | SELECT pg_sleep(10);
```

I would love some feedback on what column to go for (is `usename` (what postgres defines) fine?).

### DIFF
--- a/commands/ps.js
+++ b/commands/ps.js
@@ -12,6 +12,7 @@ function * run (context, heroku) {
   const {verbose} = flags
 
   let db = yield fetcher.database(app, database)
+  let addon = yield fetcher.addon(app, database)
 
   const num = Math.random()
   const waitingMarker = `${num}${num}`
@@ -27,20 +28,18 @@ SELECT '${num}' || '${num}' WHERE EXISTS (
   let waiting = waitingOutput.includes(waitingMarker)
                 ? 'waiting'
                 : 'wait_event IS NOT NULL AS waiting'
-
   let query = `
 SELECT
  pid,
  state,
  application_name AS source,
+ usename AS username,
  age(now(),xact_start) AS running_for,
  ${waiting},
  query
 FROM pg_stat_activity
 WHERE
- query <> '<insufficient privilege>'
- ${verbose ? '' : "AND state <> 'idle'"}
- AND pid <> pg_backend_pid()
+ pid <> pg_backend_pid()
  ORDER BY query_start DESC
 `
 
--- a/test/commands/ps.js
+++ b/test/commands/ps.js
@@ -44,6 +44,7 @@ describe('pg:ps', () => {
  pid,
  state,
  application_name AS source,
+ usename,
  age(now(),xact_start) AS running_for,
  wait_event IS NOT NULL AS waiting,
  query
@@ -61,6 +62,7 @@ WHERE
  pid,
  state,
  application_name AS source,
+ usename,
  age(now(),xact_start) AS running_for,
  wait_event IS NOT NULL AS waiting,
  query


msakrejda:
What problem are you trying to solve?

keiko713:
I think `username` is better than `usename`, we're changing `application_name` to `source` so we can change it.
@uhoh-itsmaciek I _think_ this is about credentials. Each `usename` will be the name of credentials (except the default one), and I guess it's nice to see it in `heroku pg:ps`?

msakrejda:
Ok, got it. I don't suppose we can try to flag the original role as the default credential? It's unfortunate that we refer to it as "default" everywhere but it will show up as something autogenerated and opaque here...

msakrejda:
On second thought, I guess `pg:ps` and other similar commands are thin wrappers around useful psql queries right now. I think we should stop relying on `psql`, change them to use a real node Postgres driver with no native deps (so they work even without `psql` installed), and present the output in a more user-friendly way that's idiomatic with the general output style of other CLI commands. But that's obviously a bit of a project, and way beyond the scope of this patch.

So, I think we basically have two options here: either we ship this more or less as is (`usename` or `username` or `user` are all fine), or we try to enhance this with an indicator of the default role. We could do the latter by calling the API to check the default role name and then adding something like `usename = ${defaultRole} AS default` to the query.

camillebaldock:
@uhoh-itsmaciek yes this is indeed about credentials and seeing which credential is doing what queries (i am planning to make similar changes to pg extras too, to give users visibility on what user is creating for example a long running/blocking query); i think using "Default" at the moment in the usename at the moment could confuse people since credentials is still private beta; it is however a good suggestion; i'm therefore going to make the change to mention "Default" here but flag this PR as only-for-GA so that we don't cause confusion before credentials + documentation is available to all our users -> would that work for you?

keiko713:
I'm +1 for what Camille said about "default", it can be confusing so maybe we want to avoid it for now.
I'd still like to see `username` instead of `usename`, this column name is something I got confused when I first started looking at `pg_stat_activity` table and given that `heroku pg:ps` is used by many I'm-not-familiar-with-postgres users, saying `username` would help I believe.

msakrejda:
Yup, sounds reasonable.

camillebaldock:
@uhoh-itsmaciek @keiko713 i'm now rethinking this as I think it needs a bit more thought. There's a gotcha with the fact that `pg:ps` does not include queries on which it has insufficient privilege -> this hides things from postgres/collectd which is correct, but we shouldn't be hiding things from a credential they have misconfigured with a blocking query...

camillebaldock:
Closing this and rethinking (See comment above). there will an email.

msakrejda:
Hmm, yeah, it looks like very little info is actually exposed:

```
bob=> select * from pg_stat_activity;
-[ RECORD 1 ]----+--------------------------------
datid            | 2830928
datname          | bob
pid              | 22523
usesysid         | 16384
usename          | maciek
application_name | psql
client_addr      | 💩
client_hostname  | 💩
client_port      | 💩
backend_start    | 💩
xact_start       | 💩
query_start      | 💩
state_change     | 💩
waiting          | 💩
state            | 💩
backend_xid      | 💩
backend_xmin     | 24022227
query            | <insufficient privilege>
```
Not having xact_start here kinda sucks.

PR #124, opened 2017-06-01
Include username and transaction start in pg:ps output

camillebaldock:
Follow-up from https://github.com/heroku/heroku-pg/pull/122.


msakrejda (approved):
+1

PR #125, opened 2017-06-01
remove existing sigint listeners

jdx:
note that running v6 without having it installed will cancel when running ctrl-c no matter what (seems to be something with `babel-node`) so to test this you'll need a fully installed version of v6


msakrejda (approved):
Seems reasonable.

PR #128, opened 2017-06-05
Use highest version of heroku-cli-addons

camillebaldock:
Releasing this so we can fix https://trello.com/c/c6VHpmy1/2014-%5Bcreds%5D-cannot-list-credentials-and-specify-a-database-if-a-database-has-over-two-attachments-for-the-same-app.
 
This version includes the fix done in https://github.com/heroku/heroku-cli-addons/pull/80 so we can solve ambiguous resolution of postgres addons.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "co-wait": "0.0.0",
     "debug": "2.6.3",
     "filesize": "3.5.6",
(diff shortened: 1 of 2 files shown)


msakrejda (approved):
+1

PR #130, opened 2017-06-07
Display additional rotation details

camillebaldock:
Displays connection information during additional credential rotation

screen shot 2017-06-07 at 11 54 37 am

### DIFF
--- a/commands/credentials.js
+++ b/commands/credentials.js
@@ -3,7 +3,7 @@
 const co = require('co')
 const cli = require('heroku-cli-util')
 const sortBy = require('lodash.sortby')
-const printf = require('printf')
+const util = require('../lib/util')
 
 function * run (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
@@ -16,22 +16,6 @@ function * run (context, heroku) {
     let attachments = []
     let credentials = []
 
-    function formatAttachment (attachment) {
-      let attName = cli.color.addon(attachment.name)
-
-      let output = [cli.color.dim('as'), attName]
-      let appInfo = `on ${cli.color.app(attachment.app.name)} app`
-      output.push(cli.color.dim(appInfo))
-
-      return output.join(' ')
-    }
-
-    function renderAttachment (attachment, app, isLast) {
-      let line = isLast ? '└─' : '├─'
-      let attName = formatAttachment(attachment)
-      return printf(' %s %s', cli.color.dim(line), attName)
-    }
-
     function presentCredential (cred) {
       let credAttachments = []
       if (cred !== 'default') {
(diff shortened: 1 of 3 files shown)


commands/credentials.js
```diff
@@ -51,11 +52,36 @@ function * run (context, heroku) {
...
+      let credentialStore = credentials.filter(a => a.name === cred)[0]
+      if (credentialStore.state === 'rotating') {
+        let formatted = credentialStore.credentials.map(function (credential, idx) {
+          return {
+            'user': credential.user,
+            'state': credential.state,
+            'connections': credential.connections
+          }
+        })
+        let connectionInformationAvailable = formatted.some(function (c) { return c.connections })
```
msakrejda: Not a big deal, but I find something like `formatted.some((c) => c.connections)` more readable...

commands/credentials.js
```diff
@@ -51,11 +52,36 @@ function * run (context, heroku) {
...
+        if (connectionInformationAvailable) {
+          let prefix = '       '
+          rotationLines.push(`${prefix}Usernames currently active for this credential:`)
+          cli.table(formatted, {
+            printHeader: false,
+            printLine: function (line) { rotationLines.push(line) },
+            columns: [
+              {key: 'user', format: (v, r) => `${prefix}${v}`},
+              {key: 'state', format: (v, r) => (v === 'revoking') ? 'waiting for no connections to be revoked' : v},
```
msakrejda: `'waiting for no connections to be revoked'`?

commands/credentials.js
```diff
@@ -51,11 +52,36 @@ function * run (context, heroku) {
...
+        let formatted = credentialStore.credentials.map(function (credential, idx) {
+          return {
+            'user': credential.user,
+            'state': credential.state,
+            'connections': credential.connections
+          }
+        })
+        let connectionInformationAvailable = formatted.some(function (c) { return c.connections })
+        if (connectionInformationAvailable) {
+          let prefix = '       '
```
msakrejda: This seems like a pretty brittle way to handle this--if the ` ├─ as ` formatting changes at all, this will break. On the other hand, I don't have a great suggestion; maybe this is the best we've got for now.

commands/credentials.js
```diff
@@ -51,11 +52,36 @@ function * run (context, heroku) {
...
+            printLine: function (line) { rotationLines.push(line) },
+            columns: [
+              {key: 'user', format: (v, r) => `${prefix}${v}`},
+              {key: 'state', format: (v, r) => (v === 'revoking') ? 'waiting for no connections to be revoked' : v},
+              {key: 'connections', format: (v, r) => `${v} connections`}
+            ]
+          })
+        }
+      }
+      return [cred].concat(attLines).concat(rotationLines).join('\n')
```
msakrejda: Can we hoist these helper functions out of `run`? It bloats run (just in terms of lines of code) and makes scope issues more confusing.

msakrejda: But it can wait till another PR.

msakrejda (approved):
+1

commands/credentials.js
camillebaldock: yes :-S i want to express the fact that it is in the rotation process at the point where we are waiting for the credential to have no connections before revoking it. any copy suggestion would be super welcome ❤️

msakrejda: Ah, I totally misunderstood the message. I got confused so I think there's room for improvement, but I'm not sure about the copy. I think we can ship as is and revise...

PR #131, opened 2017-06-08
Valid options: only return usable database URLs

camillebaldock:
Given that pg commands do not work when requesting a credential URL, do not suggest the credential URL as a valid options when suggesting what could be used in the command.


msakrejda (approved):
+1

PR #136, opened 2017-06-22
example jest migration

jdx:


msakrejda:
It's great to have this as an example, but I don't think we're going to merge this or make the change anytime son, so I'm just going to close for now. We can always refer to it later.

PR #138, opened 2017-06-27
Use fuzzy matching for looking up forks/followers in pg:info

msakrejda:
See https://github.com/heroku/shogun/pull/7712

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -5,6 +5,7 @@ const debug = require('./debug')
 const getBastion = require('./bastion').getBastion
 const sortBy = require('lodash.sortby')
 const printf = require('printf')
+const URL = require('url').URL
 
 function getUrl (configVars) {
   let connstringVars = configVars.filter((cv) => (cv.endsWith('_URL')))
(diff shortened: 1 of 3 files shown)


msakrejda:
@camilleldn is there anything in the credentials world that this change would cause problems for?

lib/util.js
```diff
@@ -5,6 +5,7 @@ const debug = require('./debug')
 const getBastion = require('./bastion').getBastion
 const sortBy = require('lodash.sortby')
 const printf = require('printf')
+const URL = require('url').URL
```
jdx, line 8: TIL url.parse is legacy

jdx:
code lgtm aside from test failure

jdx (approved):

msakrejda:
Hmm, I'm not sure what's up with the failure: tests pass locally. Investigating.

msakrejda:
The issue was due to 6.2.2 behaving differently. I did not dig in since the CLI is on a more recent version.

camillebaldock (approved):
👍 this will work well with credentials

jdx:
I think that `new URL` syntax is new

PR #141, opened 2017-07-18
Add pg:settings command

cyberdelia:
Add pg:settings command, mostly following the design spec.

### DIFF
--- a/commands/settings/index.js
+++ b/commands/settings/index.js
@@ -1,0 +1,31 @@
+'use strict'
+
+const cli = require('heroku-cli-util')
+const co = require('co')
+
+function * run (context, heroku) {
+  const host = require('../../lib/host')
+  const util = require('../../lib/util')
+  const fetcher = require('../../lib/fetcher')(heroku)
+  const {app, args} = context
+
+  const db = yield fetcher.addon(app, args.database)
+
+  if (util.starterPlan(db)) throw new Error('This operation is not supported by Hobby tier databases.')
+  if (util.legacyPlan(db)) throw new Error('This operation is not supported by Legacy-tier databases.')
+
+  let settings = yield heroku.get(`/postgres/v0/databases/${db.id}/config`, {host: host(db)})
+  cli.styledHeader(db.name)
+  Object.keys(settings).forEach((key) => { settings[key] = settings[key]['value'] })
+  cli.styledObject(settings)
+}
+
+module.exports = {
+  topic: 'pg',
+  command: 'settings',
+  description: 'show current settings',
+  needsApp: true,
+  needsAuth: true,
+  args: [{name: 'database', optional: true}],
+  run: cli.command({preauth: true}, co.wrap(run))
+}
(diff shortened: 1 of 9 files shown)


commands/settings/index.js
```diff
@@ -0,0 +1,31 @@
...
+function * run (context, heroku) {
+  const host = require('../../lib/host')
+  const util = require('../../lib/util')
+  const fetcher = require('../../lib/fetcher')(heroku)
+  const {app, args} = context
+
+  const db = yield fetcher.addon(app, args.database)
+
+  if (util.starterPlan(db)) throw new Error('This operation is not supported by Hobby tier databases.')
+  if (util.legacyPlan(db)) throw new Error('This operation is not supported by Legacy-tier databases.')
```
msakrejda, line 15: Legacy's not really a "tier", so maybe just "not supported for legacy database plans"? Also, should we link to [Dev Center](https://devcenter.heroku.com/articles/heroku-postgres-legacy-plans)?

commands/settings/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+  let settings = yield heroku.get(`/postgres/v0/databases/${db.id}/config`, {host: host(db)})
+  cli.styledHeader(db.name)
+  Object.keys(settings).forEach((key) => { settings[key] = settings[key]['value'] })
```
msakrejda: Can you avoid mutating the object you are traversing to minimize chance of shenanigans in the future? Something like

```javascript
let remapped = Object.keys(settings).reduce((s, key) => { s[key] = settings[key]['value']; return s }, {})
```

commands/settings/log_lock_waits.js
```diff
@@ -0,0 +1,23 @@
...
+function explain (setting) {
+  if (setting.value) {
+    return `When a deadlock is detected, a log message will be emitted in your application's logs.`
+  }
+  return `When a deadlock is detected, no log message will be emitted in your application's logs.`
+}
+
+module.exports = {
+  topic: 'pg',
+  command: 'settings:log_lock_waits',
```
msakrejda: @dickeyxxx how do you feel about introducing underscored commands? On the one hand, this is consistent with the postgres setting, but it seems other core commands use kebab-case, no? I think consistency with the CLI conventions here is more important, but I could be convinced.

lib/setter.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+exports.enum = (value) => {
+  return value
+}
+
+exports.numeric = (value) => {
+  let n = Number(value)
+  if (isNaN(n)) {
```
msakrejda: Maybe `if (!isFinite(n)) {` instead? I doubt we have much use for an `Infinity` or `-Infinity` value for settings, which this still allows.

commands/settings/log_lock_waits.js
jdx: we try to avoid it if all possible, but yeah, kebab-case is what the style guide specifies

commands/settings/index.js
cyberdelia: Paging @neovintage

msakrejda: Also either both "Hobby tier" and "Legacy-tier" should be hyphenated, or neither should be. And for what it's worth, I'm fine shipping after making that consistent, and then sorting the tier thing out later if necessary.

msakrejda (approved):
+1

PR #142, opened 2017-07-31
Small improvements to pg:settings

cyberdelia:
Add support for some more various booleans, and a slightly better description.

### DIFF
--- a/commands/settings/index.js
+++ b/commands/settings/index.js
@@ -17,7 +17,7 @@ function * run (context, heroku) {
   let settings = yield heroku.get(`/postgres/v0/databases/${db.id}/config`, {host: host(db)})
   cli.styledHeader(db.name)
   let remapped = Object.keys(settings).reduce((s, key) => {
-    s[key] = settings[key]['value']
+    s[key.replace(/_/g, '-')] = settings[key]['value']
     return s
   }, {})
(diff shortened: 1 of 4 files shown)


commands/settings/index.js
```diff
@@ -26,7 +26,7 @@ function * run (context, heroku) {
 module.exports = {
   topic: 'pg',
   command: 'settings',
-  description: 'show current settings',
+  description: 'shows your database current postgres settings values',
```
msakrejda: Comparing to other CLI addons, this should describe the action, rather than the command, so I believe this should stay `show`. And `values` is redundant here and it's pretty obvious that we're talking about `postgres`, so maybe something like:

> show your current database settings

?

cyberdelia: Paging @TheBecky.

P.S: @uhoh-itsmaciek usually wins the Shakespeare award.

bexhearty: Yes! @uhoh-itsmaciek won the Shakespeare award

> show your current database settings

msakrejda: Shall I compare thee to `log_lock_waits`? Thou art more lovely and more temperate.

msakrejda (approved):
+1

PR #150, opened 2017-08-08
Allow for default credential to be forced with a warning

genslein:
Given the new functionality and flexibility desired with Credentials as a whole, this is to allow maintenances for DBs to run more smoothly and help expedite those types of maintenance works as well.

### DIFF
--- a/commands/credentials/rotate.js
+++ b/commands/credentials/rotate.js
@@ -10,27 +10,23 @@ function * run (context, heroku) {
   const {app, args, flags} = context
   let db = yield fetcher.addon(app, args.database)
   let all = flags.all
+  let warnings = []
 
   if (all && flags.name !== undefined) {
     throw new Error('cannot pass both --all and --name')
   }
-  let cred = flags.name || 'default'
-  if ((cred === 'default' || all) && flags.force) {
-    if (all) {
-      throw new Error('Cannot force rotate all credentials: the default credential cannot be force rotated.')
-    } else {
-      throw new Error('Cannot force rotate the default credential.')
-    }
-  }
   if (util.starterPlan(db) && cred !== 'default') {
     throw new Error(`Only one default credential is supported for Hobby tier databases.`)
+  }
+  let cred = flags.name || 'default'
+  if (all && flags.force) {
+    warnings.push('WARNING: This forces rotation on all credentials including the default credential.')
   }
   let attachments = yield heroku.get(`/addons/${db.name}/addon-attachments`)
   if (flags.name) {
     attachments = attachments.filter(a => a.namespace === `credential:${cred}`)
   }
 
-  let warnings = []
   if (!flags.all) {
     warnings.push(`The password for the ${cred} credential will rotate.`)
   }
--- a/commands/credentials/rotate.js
+++ b/commands/credentials/rotate.js
@@ -20,7 +20,7 @@ function * run (context, heroku) {
     throw new Error(`Only one default credential is supported for Hobby tier databases.`)
   }
   if (all && flags.force) {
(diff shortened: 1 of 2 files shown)


commands/credentials/rotate.js
```diff
@@ -10,27 +10,23 @@ function * run (context, heroku) {
...
-    } else {
-      throw new Error('Cannot force rotate the default credential.')
-    }
-  }
   if (util.starterPlan(db) && cred !== 'default') {
     throw new Error(`Only one default credential is supported for Hobby tier databases.`)
   }
+  let cred = flags.name || 'default'
+  if (all && flags.force) {
+    warnings.push('WARNING: This forces rotation on all credentials including the default credential.')
```
msakrejda: The other warnings don't start with `WARNING: `; we should be consistent.

test/commands/credentials/rotate.js
```diff
@@ -100,16 +100,6 @@ describe('pg:credentials:rotate', () => {
     return expect(cmd.run({app: 'myapp', args: {}, flags: {all: true, name: 'my_role', confirm: 'myapp'}}), 'to be rejected with', err)
   })
 
-  it('fails with an error if both --force and --all are included', () => {
-    const err = new Error(`Cannot force rotate all credentials: the default credential cannot be force rotated.`)
-    return expect(cmd.run({app: 'myapp', args: {}, flags: {force: true, all: true, confirm: 'myapp'}}), 'to be rejected with', err)
-  })
```
msakrejda, line 106: We should assert on the expected warning rather than drop this test, no?

msakrejda (changes requested):
Thanks, some minor comments, but looks good.

We should also change the time in https://github.com/heroku/heroku-pg/blob/6a8122a6b13b1f1c2d2eb7ce3777e39ee7bebe93/commands/credentials/rotate.js#L36 , no?

test/commands/credentials/rotate.js
genslein: @uhoh-itsmaciek we don't use a temporary username for default so that wouldn't be accurate. Also that seems like more information than we should give? That only applies to additional credentials.

genslein:
@uhoh-itsmaciek we don't use a temporary username for default so that wouldn't be accurate. Also that seems like more information than we should give? That only applies to additional credentials.

test/commands/credentials/rotate.js
genslein: adding the test  :)

genslein:
@uhoh-itsmaciek ready for a 2nd pass

commands/credentials/rotate.js
```diff
@@ -10,27 +10,23 @@ function * run (context, heroku) {
...
-      throw new Error('Cannot force rotate all credentials: the default credential cannot be force rotated.')
-    } else {
-      throw new Error('Cannot force rotate the default credential.')
-    }
-  }
   if (util.starterPlan(db) && cred !== 'default') {
     throw new Error(`Only one default credential is supported for Hobby tier databases.`)
   }
+  if (all && flags.force) {
+    warnings.push('WARNING: This forces rotation on all credentials including the default credential.')
```
msakrejda: I think [this comment](https://github.com/heroku/heroku-pg/pull/150#discussion_r132044716) got lost amid your other changes, but other than that, this looks good.

msakrejda (changes requested):

camillebaldock (approved):

commands/credentials/rotate.js
genslein: I did it originally to warn of the footgun here but will remove for style consistency.

msakrejda (approved):
:+1:

PR #151, opened 2017-08-17
show query under debug

jdx:
example:

```
$ DEBUG=pg heroku pg:kill -a forker 1
  pg fetching DATABASE_URL on forker +0ms
  pg Using "DATABASE_URL" to connect to your database… +2s
  pg Running query: SELECT pg_cancel_backend(1); +3ms
WARNING:  PID 1 is not a PostgreSQL server process
 pg_cancel_backend
-------------------
 f
(1 row)
```

### DIFF
--- a/lib/psql.js
+++ b/lib/psql.js
@@ -18,6 +18,7 @@ function execPsql (query, dbEnv) {
   const {spawn} = require('child_process')
   return new Promise((resolve, reject) => {
     let result = ''
+    debug('Running query: %s', query.trim())
     let psql = spawn('psql', ['-c', query], {env: dbEnv, encoding: 'utf8', stdio: [ 'ignore', 'pipe', 'inherit' ]})
     psql.stdout.on('data', function (data) {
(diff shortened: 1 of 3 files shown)


lib/psql.js
```diff
@@ -51,6 +52,7 @@ function * exec (db, query) {
   let configs = bastion.getConfigs(db)
 
   yield bastion.sshTunnel(db, configs.dbTunnelConfig)
+  debug('Running query: %s', query.trim())
```
msakrejda: Curious why not push down into `execPsql`? It's only called from here currently, so it's moot, but I guess since that's a lower-level path we may want to use it for things that we don't want to log at some point?

msakrejda (approved):

lib/psql.js
jdx: moved it down, my original thought was that execPsql already looked a bit overloaded and could maybe be broken up into a couple functions, but on closer look it seems fine

PR #154, opened 2017-09-01
Promote: make it work with credentials

camillebaldock:
pg:promote could not be used with a credential currently and trying to do so fails with an unhelpful error message, or appeared like it had worked but had actually attached the default credential of that addon. 

This PR corrects that. 

https://trello.com/c/D2wFNu3p/2223-pgpromote-make-it-work-with-credentials

### DIFF
--- a/commands/promote.js
+++ b/commands/promote.js
@@ -17,8 +17,6 @@ function * run (context, heroku) {
     // If no current DATABASE, return nil.
     let attachments = yield heroku.get(`/apps/${app}/addon-attachments`)
     let current = attachments.find(a => a.name === 'DATABASE')
-    console.log(current)
-    console.log(attachment)
     if (!current) return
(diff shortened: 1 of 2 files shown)


commands/promote.js
```diff
@@ -17,9 +17,17 @@ function * run (context, heroku) {
     // If no current DATABASE, return nil.
     let attachments = yield heroku.get(`/apps/${app}/addon-attachments`)
     let current = attachments.find(a => a.name === 'DATABASE')
+    console.log(current)
+    console.log(attachment)
```
msakrejda: Can you drop the debugging statements (or change them to use `debug`)?

camillebaldock: removed, i had forgotten to remove those at the end of development

msakrejda (approved):
Thanks!

PR #157, opened 2017-09-30
Capture snapshot

camillebaldock:
Behind private beta flag for now

### DIFF
--- a/commands/backups/capture.js
+++ b/commands/backups/capture.js
@@ -12,22 +12,27 @@ function * run (context, heroku) {
   const interval = Math.max(3, parseInt(flags['wait-interval'])) || 3
   const db = yield fetcher.addon(app, args.database)
 
-  let backup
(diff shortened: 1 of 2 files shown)


keiko713:
During today's demo, we got a feedback that we should show the timestamp when the snapshot was taken (and maybe snapshot id).

```
$ heroku pg:backups:capture HEROKU_POSTGRESQL_KEIKO_URL --snapshot -a ck-snapshot-test
Taking snapshot of postgresql-keiko-objective-12233... done
```

Well actually, it's more like shogun side change than CLI side change?

camillebaldock:
@keiko713 yes i think there will be a change needed here is show the snapshot id + timestamp from shogun. We will need more copy here anyway to explain what a snapshot is. There is also the fact that down the line we won't do this through --snapshot but directly through pg:backups:capture so I'd rather we address what the exact return should be as a separate PR.

keiko713 (approved):
Ok I'm fine with that

PR #158, opened 2017-10-02
Add confirmation and warning for DBs with continuous protection

coreypurcell:
If a database has continuous protection, then we warn the user that logical backups are not
necessary and that backups of large databases may timeout.

### DIFF
--- a/commands/backups/capture.js
+++ b/commands/backups/capture.js
@@ -24,7 +24,7 @@ function * run (context, heroku) {
   if (dbInfo) {
     let dbProtected = /On/.test(dbInfo.info.find(attribute => attribute.name === 'Continuous Protection').values[0])
     if (dbProtected) {
-      yield cli.confirmApp(app, flags.confirm, 'Continuous protection is already enabled for this database. Logical backups of large databases are likely to timeout. See https://devcenter.heroku.com/articles/heroku-postgres-data-safety-and-continuous-protection#physical-backups-on-heroku-postgres.  Are you sure?')
+      cli.warn(app, flags.confirm, 'Continuous protection is already enabled for this database. Logical backups of large databases are likely to fail. See https://devcenter.heroku.com/articles/heroku-postgres-data-safety-and-continuous-protection#physical-backups-on-heroku-postgres.')
     }
   }
 
--- a/commands/backups/capture.js
+++ b/commands/backups/capture.js
@@ -25,7 +25,6 @@ function * run (context, heroku) {
       if (err.statusCode !== 404) throw err
       cli.exit(1, `${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
     })
-    console.log(dbInfo.info)
     if (dbInfo) {
       let dbProtected = /On/.test(dbInfo.info.find(attribute => attribute.name === 'Continuous Protection').values[0])
       if (dbProtected) {
--- a/commands/backups/capture.js
+++ b/commands/backups/capture.js
@@ -12,6 +12,22 @@ function * run (context, heroku) {
   const interval = Math.max(3, parseInt(flags['wait-interval'])) || 3
   const db = yield fetcher.addon(app, args.database)
 
+  let dbInfo = yield heroku.request({
+    host: host(db),
+    method: 'get',
+    path: `/client/v11/databases/${db.id}`
+  }).catch(err => {
+    if (err.statusCode !== 404) throw err
+    cli.warn(`${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
+  })
+
+  if (dbInfo) {
+    let dbProtected = /On/.test(dbInfo.info.find(attribute => attribute.name === 'Continuous Protection').values[0])
+    if (dbProtected) {
+      yield cli.confirmApp(app, flags.confirm, 'Continuous protection is already enabled for this database. Logical backups of large databases are likely to timeout. Are you sure?')
+    }
+  }
+
   let backup
   yield cli.action(`Starting backup of ${cli.color.addon(db.name)}`, co(function * () {
     backup = yield heroku.post(`/client/v11/databases/${db.id}/backups`, {host: host(db)})
@@ -39,7 +55,8 @@ module.exports = {
   args: [{name: 'database', optional: true}],
   flags: [
     {name: 'wait-interval', hasValue: true},
-    {name: 'verbose', char: 'v'}
+    {name: 'verbose', char: 'v'},
+    {name: 'confirm', char: 'c', hasValue: true}
   ],
   run: cli.command({preauth: true}, co.wrap(run))
 }
--- a/commands/backups/capture.js
+++ b/commands/backups/capture.js
@@ -18,13 +18,13 @@ function * run (context, heroku) {
     path: `/client/v11/databases/${db.id}`
   }).catch(err => {
     if (err.statusCode !== 404) throw err
-    cli.warn(`${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
+    cli.exit(1, `${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
   })
 
   if (dbInfo) {
     let dbProtected = /On/.test(dbInfo.info.find(attribute => attribute.name === 'Continuous Protection').values[0])
     if (dbProtected) {
-      yield cli.confirmApp(app, flags.confirm, 'Continuous protection is already enabled for this database. Logical backups of large databases are likely to timeout. Are you sure?')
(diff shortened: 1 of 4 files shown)


coreypurcell:
open to improvements, both in the wording of the warning and my bad javascripting abilities.

commands/backups/capture.js
```diff
@@ -12,6 +12,22 @@ function * run (context, heroku) {
   const interval = Math.max(3, parseInt(flags['wait-interval'])) || 3
   const db = yield fetcher.addon(app, args.database)
 
+  let dbInfo = yield heroku.request({
+    host: host(db),
+    method: 'get',
+    path: `/client/v11/databases/${db.id}`
+  }).catch(err => {
+    if (err.statusCode !== 404) throw err
+    cli.warn(`${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
```
msakrejda: It seems weird to warn that the DB is not yet provisioned but then allow you to try to take a backup anyway. Are there any cases in which that can succeed?

commands/backups/capture.js
```diff
@@ -12,6 +12,22 @@ function * run (context, heroku) {
...
+  })
+
+  if (dbInfo) {
+    let dbProtected = /On/.test(dbInfo.info.find(attribute => attribute.name === 'Continuous Protection').values[0])
+    if (dbProtected) {
+      yield cli.confirmApp(app, flags.confirm, 'Continuous protection is already enabled for this database. Logical backups of large databases are likely to timeout. Are you sure?')
```
msakrejda: Should we link to Dev Center here?

msakrejda:
@coreypurcell @caseman @TheBecky it occurs to me that requiring `--confirm` on `pg:backups:capture` will break users who have scripted the CLI (instead of relying on backup scheduling for whatever reason). Should we try to figure out how many users might be impacted?

commands/backups/capture.js
coreypurcell: I don't know, but that seems unlikely to ever succeed. Maybe just bail out at that point?

msakrejda: 👍

keiko713:
@uhoh-itsmaciek ah, good point, haven't thought about it. It can indeed break some customer's cron job or whatever. I at least know a few customers who do that kind of terrible things.
We.. may just want to add a warning instead of doing `--confirm` thing?

keiko713:
There are 12 apps that took manual backups more than 10 times last 8 hours. Very likely the customer who has some sort of automated jobs. When I change condition to `count(group_id) > 5`, the result was 73 🤦‍♂️ 

```
(47324) /d3ae9ld53pnob0=> select group_id, count(group_id) as count from transfers where created_at > '2017-10-02 12:00:00' and schedule_id is null group by group_id having count(group_id) > 10;
               group_id               │ count 
──────────────────────────────────────┼───────
 050680ce-ccec-4748-9a2f-5868fd8eb22b │    60
 d32669c5-a316-47be-a422-6a61c079f855 │    19
 33d3ef87-b792-4902-95d6-7d94ce4c8a2a │    60
 4d601dae-7c88-46e2-b55f-675c6fadb395 │    14
 9505d881-5bf9-4be1-9d01-88a94afe57e0 │    53
 14e37e9b-9dfc-4a5e-93d4-3366e6c9086e │    13
 d2903cce-b24a-4d23-be59-55b6e40e04c3 │    18
 59927e3e-fc9f-4a76-a573-7e8a25031fd9 │    17
 2cc1929a-f00e-4cfb-b15f-22e1554802cf │    46
 55363752-389f-44d0-8c8e-692cbc1f6f98 │    53
 021ffdde-36c4-414c-ab64-8cbc74104ee9 │    14
 5d495ca1-feea-4784-91d6-37d7dcb09ee8 │    26
(12 rows)
```

coreypurcell:
@keiko713 @uhoh-itsmaciek so should I remove the confirmation and just output the warning?

keiko713:
@coreypurcell let's remove confirmation for now, and just output the warning. we can always revisit adding confirmation later.

commands/backups/capture.js
```diff
@@ -12,6 +12,22 @@ function * run (context, heroku) {
...
+    path: `/client/v11/databases/${db.id}`
+  }).catch(err => {
+    if (err.statusCode !== 404) throw err
+    cli.exit(1, `${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
+  })
+
+  if (dbInfo) {
+    let dbProtected = /On/.test(dbInfo.info.find(attribute => attribute.name === 'Continuous Protection').values[0])
+    if (dbProtected) {
+      yield cli.confirmApp(app, flags.confirm, 'Continuous protection is already enabled for this database. Logical backups of large databases are likely to timeout. See https://devcenter.heroku.com/articles/heroku-postgres-data-safety-and-continuous-protection#physical-backups-on-heroku-postgres.  Are you sure?')
```
keiko713: instead of "likely to timeout", can we mention something like it can result to fail... uhmm due to errors? or "likely to fail"?
we're seeing two issues with the large database: 1) timing out because taking a logical backup can't finish within 24 hours 2) some OOM on either dyno side or postgres side or something that just impossible to take a backup. While both of them are the failure, actually for the case of 1), you don't really see "timeout" as the result (to the user, it either marked as "failed" without obvious reason or the state doesn't change from "running"), so I thought that mentioning "timeout" can be confusing.

coreypurcell:
@keiko713 should be ready for another review, switched the wording and removed the confirmation.

msakrejda (changes requested):
Looks great. Can you add tests for this functionality?

commands/backups/capture.js
```diff
@@ -25,10 +25,11 @@ function * run (context, heroku) {
       if (err.statusCode !== 404) throw err
       cli.exit(1, `${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
     })
+    console.log(dbInfo.info)
```
msakrejda: Stray debugging line?

test/commands/backups/schedule.js
```diff
@@ -40,13 +38,36 @@ const shouldSchedule = function (cmdRun) {
...
+
+    return cmdRun({app: 'myapp', args: {}, flags: {at: '06:00 EDT'}})
+    .then(() => expect(cli.stderr, 'to match', /backups of large databases are likely to fail/))
+  })
+
+  it('does not warn user that logical backups are error prone if continuous proctecion is off', () => {
+    let dbA = {info: [
+      {name: 'Continuous Protection', values: ['Off']}
+    ]}
+    // pg = nock('https://postgres-api.heroku.com')
```
msakrejda: Can you drop the commented-out code?

commands/backups/capture.js
```diff
@@ -17,6 +17,21 @@ function * run (context, heroku) {
...
+      path: `/client/v11/databases/${db.id}`
+    }).catch(err => {
+      if (err.statusCode !== 404) throw err
+      cli.exit(1, `${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
+    })
+    console.log(dbInfo.info)
+    if (dbInfo) {
+      let dbProtected = /On/.test(dbInfo.info.find(attribute => attribute.name === 'Continuous Protection').values[0])
+      if (dbProtected) {
+        cli.warn('Continuous protection is already enabled for this database. Logical backups of large databases are likely to fail. See https://devcenter.heroku.com/articles/heroku-postgres-data-safety-and-continuous-protection#physical-backups-on-heroku-postgres.')
```
keiko713: let's add new line between `fail.` and `See` (I think just adding `\n` is fine? @uhoh-itsmaciek may know more)

keiko713:
overall looks good but please address Maciek's review. I'll take a look again once you fixed them :)

commands/backups/capture.js
msakrejda: Yeah, that'd work. Or you can do two separate `cli.warn`s.

commands/backups/capture.js
```diff
@@ -17,6 +17,21 @@ function * run (context, heroku) {
...
   } else {
+    let dbInfo = yield heroku.request({
+      host: host(db),
+      method: 'get',
+      path: `/client/v11/databases/${db.id}`
+    }).catch(err => {
+      if (err.statusCode !== 404) throw err
+      cli.exit(1, `${cli.color.addon(db.name)} is not yet provisioned.\nRun ${cli.color.cmd('heroku addons:wait')} to wait until the db is provisioned.`)
+    })
+    console.log(dbInfo.info)
```
msakrejda: Is this intentional? Looks like [my original comment](https://github.com/heroku/heroku-pg/pull/158#discussion_r142765161) no longer shows up, but this is just debugging info, right?

coreypurcell: no I thought I removed it

coreypurcell: ah looks like I never saved the file

keiko713 (approved):
I'm fine with this, but I think Mr. Maciek needs to approve too.

msakrejda (approved):
Awesome, thanks!

PR #160, opened 2017-10-17
Adding 'exclude' option to pg:pull/push for excluding data from given tables

BenTalagan:
Pretty straight forward feature, but some points may be discussed :)

### DIFF
--- a/commands/pull.js
+++ b/commands/pull.js
@@ -168,7 +168,7 @@ function * pull (context, heroku) {
 
   const source = yield fetcher.database(app, args.source)
   const target = parseURL(args.target)
-  
+
   cli.log(`heroku-cli: Pulling ${cli.color.addon(source.attachment.addon.name)} ---> ${cli.color.cyan(args.target)}`)
   yield run(source, target, exclusions)
   cli.log('heroku-cli: Pulling complete.')
--- a/commands/pull.js
+++ b/commands/pull.js
@@ -150,7 +150,7 @@ function * push (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   const {app, args} = context
   const flags = context.flags
-  const exclusions = parseExclusions(flags['exclude'])
+  const exclusions = parseExclusions(flags['exclude-table-data'])
 
   const source = parseURL(args.source)
   const target = yield fetcher.database(app, args.target)
@@ -164,7 +164,7 @@ function * pull (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   const {app, args} = context
   const flags = context.flags
-  const exclusions = parseExclusions(flags['exclude'])
+  const exclusions = parseExclusions(flags['exclude-table-data'])
 
   const source = yield fetcher.database(app, args.source)
   const target = parseURL(args.target)
--- a/commands/pull.js
+++ b/commands/pull.js
@@ -22,6 +22,14 @@ function parseURL (db) {
   return db
 }
 
+function parseExclusions (rawExcludeList) {
+  return (rawExcludeList || '').split(';').map(function(tname) {
+      return tname.trim();
+    }).filter(function(tname) {
+      return (tname != '');
+    });
+}
+
 function exec (cmd, opts = {}) {
   const {execSync} = require('child_process')
   debug(cmd)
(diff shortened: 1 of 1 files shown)


commands/pull.js
```diff
@@ -22,6 +22,14 @@ function parseURL (db) {
   return db
 }
 
+function parseExclusions (rawExcludeList) {
+  return (rawExcludeList || '').split(';').map(function(tname) {
+      return tname.trim();
+    }).filter(function(tname) {
+      return (tname != '');
+    });
+}
```
msakrejda, line 31: So, this is going to preclude *some* esoteric table names, but given the CLI doesn't support repeated flags (it doesn't, right @jdxcode @RasPhilCo?), that's going to happen one way or another, and most commonly-used table names are going to be supported by this.

commands/pull.js
```diff
@@ -117,14 +125,17 @@ const maybeTunnel = function * (herokuDb) {
...
 
-const run = co.wrap(function * (sourceIn, targetIn) {
+const run = co.wrap(function * (sourceIn, targetIn, exclusions) {
+  
   yield prepare(targetIn)
 
   const source = yield maybeTunnel(sourceIn)
   const target = yield maybeTunnel(targetIn)
 
+  const exclude = exclusions.map(function(e) { return '--exclude-table-data=' + e }).join(' ')
```
msakrejda, line 135: Why `--exclude-table-data` and not just `--exclude-table`? I'm not questioning the decision; I'm just curious about your use case and reasoning.

commands/pull.js
```diff
@@ -138,32 +149,39 @@ const run = co.wrap(function * (sourceIn, targetIn) {
...
 function * pull (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   const {app, args} = context
+  const flags = context.flags
+  const exclusions = parseExclusions(flags['exclude'])
 
   const source = yield fetcher.database(app, args.source)
   const target = parseURL(args.target)
-
+
```
msakrejda: Can you drop the stray whitespace here?

msakrejda (changes requested):
Looks great, thanks! There's some stray whitespace and I had a question about implementation details, but I think this is almost ready to go. Any comments @heroku/dod-ux @heroku/cli?

commands/pull.js
BenTalagan: You're right, CLI does not support repeated flags (at least that is what I could deduce from the other commands code AND documentation). So I've just followed the way it was done for other commands : same technique & separator :)

commands/pull.js
BenTalagan: Good question! My need for this feature is essentially for pulling apps databases with huge tables for local examination ; this tables contain some kind of 'advanced' logs with various criteria, so they are not crucial to the relation graph and cohesion of the database. Because pg:pull creates a new database, I would still like the table to be _created and structured_ for the app to be able to work in case I would like to perform some kind of debug. I do agree this is different from --exclude-table and options maybe should not be exclusive. In my case, I always use the heroku cli for pulling dbs it is always in the context of an application so there's always that kind of 'applicative' characterization to the operation.

msakrejda: Okay, that makes sense. We can still add the other one later if people need it. I think it's okay to expose the underlying features piecemeal in this situation.

commands/pull.js
```diff
@@ -138,32 +149,39 @@ const run = co.wrap(function * (sourceIn, targetIn) {
...
 }
 
 let cmd = {
   topic: 'pg',
   needsApp: true,
   needsAuth: true,
-  args: [{name: 'source'}, {name: 'target'}]
+  args: [{name: 'source'}, {name: 'target'}],
+  flags: [
+    {name: 'exclude', char: 'e', hasValue: true, description: 'tables for which data should be excluded (use \';\' to split multiple names)'}
```
msakrejda: What do you think about copying the `pg_dump` flag name here (`--exclude-table-data`) and dropping the short flag (since `pg_dump` doesn't have one for this option) for consistency?

commands/pull.js
BenTalagan: Agreed :)

commands/pull.js
BenTalagan: Yes I've been hesitating on this one, and have chosen a shorter version for simplicity ; but if you intend to give more options later then it's probably better to stick to pg_dump's normalization. It will make it feel more natural for people used to pg-dump, and will bring clarity to what the option really does!

msakrejda (approved):

msakrejda:
Ok, looks great to me. I'll merge tomorrow if there are no objections.

BenTalagan:
Awesome, thanks a lot!

commands/pull.js
```diff
@@ -138,32 +149,39 @@ const run = co.wrap(function * (sourceIn, targetIn) {
 function * push (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   const {app, args} = context
+  const flags = context.flags
+  const exclusions = parseExclusions(flags['exclude'])
```
msakrejda: @BenTalagan just noticed one more thing: this should have been changed to `exclude-table-data` as well in https://github.com/heroku/heroku-pg/pull/160/commits/8a780ace99682c4d67312579058cedc6537e429a right?

commands/pull.js
```diff
@@ -138,32 +149,39 @@ const run = co.wrap(function * (sourceIn, targetIn) {
...
-  yield run(source, target)
+  yield run(source, target, exclusions)
   cli.log('heroku-cli: Pushing complete.')
 }
 
 function * pull (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   const {app, args} = context
+  const flags = context.flags
+  const exclusions = parseExclusions(flags['exclude'])
```
msakrejda: And same here?

commands/pull.js
BenTalagan: Damn you're right, thank you for noticing!

commands/pull.js
BenTalagan: Ok this is fixed too.

msakrejda:
Ok, awesome, thanks. I'll merge now but wait till Monday to ship just in case. Thanks for contributing!

PR #162, opened 2017-10-19
Add pg:outliers command

cyberdelia:
Move ``pg:outliers`` command directly in the pg plugin.

### DIFF
--- a/commands/outliers.js
+++ b/commands/outliers.js
@@ -1,0 +1,61 @@
+'use strict'
+
+const co = require('co')
+const cli = require('heroku-cli-util')
+const psql = require('../lib/psql')
+
+function * run (context, heroku) {
+  const fetcher = require('../lib/fetcher')
+
+  const {app, args, flags} = context
+  const {database} = args
+
+  let db = yield fetcher(heroku).database(app, database)
+
+  if (flags.reset) {
+    yield psql.exec(db, 'SELECT pg_stat_statements_reset()')
+    return
+  }
+
+  let truncatedQueryString = flags.truncate
+    ? 'CASE WHEN length(query) <= 40 THEN query ELSE substr(query, 0, 39) || \'…\' END'
+    : 'query'
+
+  let limit = 10
+  if (context.flags.num) {
+    if (/^(\d+)$/.exec(flags.num)) {
+      limit = parseInt(flags.num)
+    } else {
+      throw new Error(`Cannot parse num param value "${flags.num}" to a number`)
+    }
+  }
+
+  let query = `
+SELECT interval '1 millisecond' * total_time AS total_exec_time,
+to_char((total_time/sum(total_time) OVER()) * 100, 'FM90D0') || '%'  AS prop_exec_time,
+to_char(calls, 'FM999G999G999G990') AS ncalls,
+interval '1 millisecond' * (blk_read_time + blk_write_time) AS sync_io_time,
+${truncatedQueryString} AS query
+FROM pg_stat_statements WHERE userid = (SELECT usesysid FROM pg_user WHERE usename = current_user LIMIT 1)
+ORDER BY total_time DESC
+LIMIT ${limit}
+`
+
+  let output = yield psql.exec(db, query)
+  process.stdout.write(output)
+}
+
+module.exports = {
+  topic: 'pg',
+  command: 'outliers',
+  description: 'show 10 queries that have longest execution time in aggregate',
+  needsApp: true,
+  needsAuth: true,
+  args: [{name: 'database', optional: true}],
+  flags: [
+    {name: 'reset', description: 'resets statistics gathered by pg_stat_statements'},
+    {name: 'truncate', char: 't', description: 'truncate queries to 40 characters'},
+    {name: 'num', char: 'n', description: 'the number of queries to display (default: 10)', hasValue: true}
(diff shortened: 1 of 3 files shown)


tom-pang (approved):

keiko713:
I think you need this Bob: https://github.com/heroku/heroku-pg-extras/blob/master/lib/util.js

cyberdelia:
@keiko713 Do we? We do install it for customer now, ok, it's async, but still.

keiko713:
At least my `keiko-sushi` doesn't have it (and lots of yobuko dbs):

```
$ heroku pg:outliers -a keiko-sushi
 ▸    pg_stat_statements extension need to be installed in the public schema first.
 ▸    This extension is only available on Postgres versions 9.2 or greater. You can install it by running:
 ▸    
 ▸    CREATE EXTENSION pg_stat_statements;
```

commands/outliers.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
+
+const co = require('co')
```
msakrejda, line 3: For what it's worth, pretty sure we can use `async/await` directly in the CLI instead of `co`, but I'm okay leaving this here for consistency (especially since it's a port of an existing command), and then porting all the commands to `async/await` in bulk.

commands/outliers.js
```diff
@@ -0,0 +1,79 @@
...
+  let query = `
+SELECT exists(
+  SELECT 1 FROM pg_extension e LEFT JOIN pg_namespace n ON n.oid = e.extnamespace
+  WHERE e.extname='pg_stat_statements' AND n.nspname = 'public'
+) AS available`
+  let output = yield psql.exec(db, query)
+
+  if (!output.includes('t')) {
+    throw new Error(`pg_stat_statements extension need to be installed in the public schema first.
+This extension is only available on Postgres versions 9.2 or greater. You can install it by running:
```
msakrejda: Fortunately, we only run Postgres version 9.2 or greater so you can drop part of this error message.

msakrejda (approved):

commands/outliers.js
keiko713: Maybe we can say something like this instead: "This extension may not be available on hobby tier Postgres."
In fact, it's not "may not be available", but I guess "is not available"?
https://trello.com/c/BK0cdQH3/2045-pg-extras-that-need-it-should-check-for-pgstatstatements-and-create-the-extension

commands/outliers.js
```diff
@@ -0,0 +1,78 @@
...
+  let query = `
+SELECT exists(
+  SELECT 1 FROM pg_extension e LEFT JOIN pg_namespace n ON n.oid = e.extnamespace
+  WHERE e.extname='pg_stat_statements' AND n.nspname = 'public'
+) AS available`
+  let output = yield psql.exec(db, query)
+
+  if (!output.includes('t')) {
+    throw new Error(`pg_stat_statements extension need to be installed in the public schema first.
+    This extension is not available on hobby tier Postgres. You can install it by running:
```
keiko713: actually this may be confusing. it's more like "this extension is not installed on hobby tier Postgres by default (but you can install if you like)", I think

cyberdelia: So I can just remove any mention of hobby then?

keiko713: That could also work maybe, at least less confusing.

gregburek:
@keiko713 does the addition of utils handle the 'pgss missing' trello card? https://trello.com/c/BK0cdQH3

commands/outliers.js
gregburek: It is 'available'. All postgres that we run, including yobukos have it as a 'shared_preload_libraries'. 

Why would we not install it automatically here?

keiko713: I'm not opposed to automatically install, but I also want to warn it may make customers confusing as it'll return "zero" result and they don't know why (well we can say why explicitly I guess).
"available" part, you're correct, I meant "may not be installed" or "is not installed".

keiko713:
@gregburek depending on what exactly you do with util. if you just do whatever it's doing now, it won't handle.

commands/outliers.js
gregburek: >This view, and the function pg_stat_statements_reset, are available only in databases they have been specifically installed into by installing the pg_stat_statements extension. However, statistics are tracked across all databases of the server whenever the pg_stat_statements module is loaded into the server, regardless of presence of the view.

So we load the library in shared_preload_libraries and install the extension in the main db. As I read the docs, it should be tracking queries and gathering statistics no matter if the extension is installed in the hobby database. A 'zero' result, then, should be true.

keiko713: oh interesting. yeah maybe worth double checking if it's correct or not with existing hobby databases. I'm gonna be PTO from tomorrow so I'll let you guys decide what to do. As long as it is not confusing to people and works, I'm fine with options that you suggested including auto-install.

msakrejda: For now, let's do the simple thing and strike `This extension is only available on Postgres versions 9.2 or greater. ` from the error message. Auto-installing is a separate issue; let's handle it in a separate patch.

msakrejda: After thinking about this some more, I'm okay with doing auto-install in this patch, too. Either works.

PR #170, opened 2017-11-30
Fix pg:pull and pg:push help messages

michaelbaudino:
The `pg:pull` help message provides two inaccurate statements:
* that TARGET must not already exist
* that TARGET will be created remotely if it's a fully qualified URL

The former is only applicable if the PG server is local and
the latter just never happens, afaics:
https://github.com/heroku/heroku-pg/blob/4ca2f4e5410efb5b77b9f51ce9379c43e8c261e8/commands/pull.js#L64-L75

The `pg:push` help message provides one inaccurate statement:
* that TARGET must not exist

It must actually exist (it's a Heroku database, could it NOT exist?)
but it must be empty.

### DIFF
--- a/commands/pull.js
+++ b/commands/pull.js
@@ -187,9 +187,9 @@ module.exports = [
   Object.assign({
     command: 'push',
     description: 'push local or remote into Heroku database',
-    help: `Push from SOURCE into TARGET. TARGET must not already exist.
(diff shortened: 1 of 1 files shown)


msakrejda:
Thank you! I've reviewed the changes and you're absolutely right. I think the discrepancies snuck in when we ported these commands from our old Ruby CLI.

msakrejda (approved):

PR #173, opened 2017-12-07
Bump node to the version currently in use in the CLI

msakrejda:
This fails tests right now, but I wanted to

 1. Open a PR to bump the version to make it clear we're drifting from the CLI and should update
 2. Ideally come up with some way of not drifting in CI, so we test against the node version our customers use

I want to address this after fixing the [current failures in CI](https://circleci.com/gh/heroku/heroku-pg/699#tests/containers/0) (that I have not been able to reproduce in running the tests locally).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,7 +7,7 @@ dependencies:
 
 machine:
   node:
-    version: v7.10.0
+    version: v9.4.0
 
 test:
   post:
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "devDependencies": {
     "heroku-client": "^3.0.2",
     "mocha": "^3.4.2",
-    "nock": "^9.0.13",
+    "nock": "9.1.6",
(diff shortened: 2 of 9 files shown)


fnando:
Maciek need to gets back to this. Soon. Eventually. Maybe.

msakrejda:
Ok, fixed and ready for review. Upgrading nock fixed the broken tests but broke twenty-some others due to some unintended changes in their mocking API and [awful error messages](https://github.com/node-nock/nock/issues/1052). I updated the broken tests.

mehtaphysical:
@uhoh-itsmaciek I fixed these tests in a different PR #180 :/

msakrejda:
Oh, sorry about that @mehtaphysical. I started working on this before you opened the issue, I think, but then yak shaved into filing the nock bug I linked above, which slowed this down. Thanks anyway!

fnando (approved):
Looks good.

PR #177, opened 2018-01-05
Add command to make a connection pooling attachment

coreypurcell:
Add command to create an attachment to a DB that uses the URL for pgbouncer. This is to avoid needing something like a build pack to create the config vars with the pgbouncer URL. 

Work to be completed:

* [x] more tests to cover the case where a credential is specified
* [x] query the api to ensure the DB actually has PGbouncer installed

### DIFF
--- a/commands/connection_pooling.js
+++ b/commands/connection_pooling.js
@@ -14,13 +14,12 @@ function * run (context, heroku) {
   let db = yield fetcher.addon(app, args.database)
   if (util.starterPlan(db)) throw new Error('This operation is not supported by Hobby tier databases.')
 
-  function createAttachment (app, as, confirm, credential) {
+  function createAttachment (app, as, credential) {
     let body = {
       name: as,
       app: {name: app},
       addon: {name: addon.name},
       namespace: `connection-pooling:${credential}`,
-      confirm
     }
     return cli.action(
       `Attaching ${credential ? cli.color.addon(credential) + ' of ' : ''}${cli.color.addon(addon.name)}${as ? ' as ' + cli.color.attachment(as) : ''} to ${cli.color.app(app)}`,
@@ -51,7 +50,7 @@ function * run (context, heroku) {
     throw new Error(`Could not find credential ${credential} with connection pooling for database ${addon.name}`)
   }
 
-  let attachment = yield util.trapConfirmationRequired(app, flags.confirm, (confirm) => createAttachment(app, flags.as, confirm, credential))
+  let attachment = yield createAttachment(app, flags.as, credential)
 
   yield cli.action(
     `Setting ${cli.color.attachment(attachment.name)} config vars and restarting ${cli.color.app(app)}`,
--- a/commands/connection_pooling.js
+++ b/commands/connection_pooling.js
@@ -11,13 +11,13 @@ function * run (context, heroku) {
 
   let db = yield fetcher.addon(app, args.database)
   let addon = yield heroku.get(`/addons/${encodeURIComponent(db.name)}`)
-  let credential = `${flags.credential || 'default'}`
+  let credential = flags.credential || 'default'
 
   if (util.starterPlan(db)) throw new Error('This operation is not supported by Hobby tier databases.')
 
   let attachment = yield cli.action(
     `Enabling Connection Pooling${credential === 'default' ? '' : ' for credential ' + cli.color.addon(credential)} on ${cli.color.addon(addon.name)} to ${cli.color.app(app)}`,
-    heroku.post(`/client/v11/databases/${db.name}/connection-pooling`, {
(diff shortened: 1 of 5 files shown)


msakrejda:
Thanks! At a glance, this looks fine, but I'ld like to give it a more thorough review. I'll try to do that by Monday.

commands/connection_pooling.js
```diff
@@ -0,0 +1,73 @@
...
+    {success: false},
+    co(function * () {
+      let releases = yield heroku.get(`/apps/${app}/releases`, {
+        partial: true,
+        headers: { 'Range': 'version ..; max=1, order=desc' }
+      })
+      cli.action.done(`done, v${releases[0].version}`)
+    })
+  )
+}
```
msakrejda, line 37: So this is largely copied from [addons:attach](https://github.com/heroku/heroku-cli-addons/blob/master/commands/addons/attach.js), right? Do you think this is the right approach @jdxcode? The code looks fine to me, but I'm nervous about the copy/pasting (though I don't really have a better suggestion at the moment).

msakrejda (commented):
Sorry about the delay. Looks reasonable though I'd love to hear from @jdxcode. Approval pending the checkbox items in the pull request description.

commands/connection_pooling.js
coreypurcell: it is very much just copied from addons:attach.

jdx: all seems fine to me!

jdx (approved):

commands/connection_pooling.js
```diff
@@ -0,0 +1,55 @@
...
+const host = require('../lib/host')
+
+function * run (context, heroku) {
+  const util = require('../lib/util')
+  const fetcher = require('../lib/fetcher')(heroku)
+  const {app, args, flags} = context
+
+  let db = yield fetcher.addon(app, args.database)
+  let addon = yield heroku.get(`/addons/${encodeURIComponent(db.name)}`)
+  let credential = `${flags.credential || 'default'}`
```
msakrejda: This works but the `flags.credentials` is going to be a string anyway, so you could do `flags.credential || 'default'` without wrapping in a template string.

commands/connection_pooling.js
```diff
@@ -0,0 +1,55 @@
...
+
+  if (util.starterPlan(db)) throw new Error('This operation is not supported by Hobby tier databases.')
+
+  let attachment = yield cli.action(
+    `Enabling Connection Pooling${credential === 'default' ? '' : ' for credential ' + cli.color.addon(credential)} on ${cli.color.addon(addon.name)} to ${cli.color.app(app)}`,
+    heroku.post(`/client/v11/databases/${db.name}/connection-pooling`, {
```
msakrejda: Why not `encodeURIComponent(db.name)` here? I mean, I don't think it really matters--I don't believe add-on names support any URL-unsafe characters--but we should be consistent.

test/commands/connection_pooling.js
```diff
@@ -0,0 +1,66 @@
...
+      pg.post(`/client/v11/databases/${addon.name}/connection-pooling`, {
+        credential: name
+      }).reply(201, {name: 'HEROKU_COLOR'})
+    })
+
+    it('attaches the pgbouncer url', () => {
+      return cmd.run({app: 'myapp', args: {database: 'postgres-1'}, flags: {credential: name}})
+      .then(() => expect(cli.stdout, 'to equal', ``))
+      .then(() => expect(cli.stderr, 'to contain', 'Setting HEROKU_COLOR config vars and restarting myapp... done, v0\n'))
+    })
```
msakrejda: Can you add tests for no credential specified so we can validate it defaults to the right thing and for the `--as` flag present, to validate that's passed along correctly?

msakrejda (changes requested):
Some minor comments, but looks great. Sorry it took me so long to follow up. Feel free to ping me in the future.

binarycleric:
Will do. Thanks for the feedback @uhoh-itsmaciek.

test/commands/connection_pooling.js
binarycleric: @uhoh-itsmaciek Done.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "heroku-pg",
   "description": "Heroku CLI plugin to manage Postgres.",
-  "version": "2.4.2",
+  "version": "2.5.0",
```
msakrejda: The version should be bumped separately and that's done automatically by `np`: see https://github.com/heroku/heroku-pg#publishing

msakrejda (changes requested):
Awesome, thanks! Just drop the `version` change and we're good to go.

msakrejda (approved):

PR #178, opened 2018-01-11
Attempt to work around issues introduced by heroku rollback

msakrejda:
The heroku rollback command can leave an app in an inconsistent state:
an add-on and attachment can be present, but the config vars may be
stripped by a rollback. Currently, the pg CLI does not handle this
correctly and will [give a cryptic error](https://support.heroku.com/tickets/544092) in this situation

This attempts to work around that by suggesting that rollback may be
at fault and how to restore the application to a sensible state (by
creating a new attachment and destroying the old one).

The new output looks like
rollback-sucks
(but without the linked plugin notice)

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -76,7 +76,13 @@ exports.presentCredentialAttachments = presentCredentialAttachments
 
 exports.getConnectionDetails = function (attachment, config) {
   const url = require('url')
-  const connstringVar = getUrl(attachment.config_vars.filter((cv) => config[cv].startsWith('postgres://')))
+  const configVars = attachment.config_vars.filter((cv) => {
+    return config[cv] && config[cv].startsWith('postgres://')
+  })
+  if (configVars.length === 0) {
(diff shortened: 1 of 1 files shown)


camillebaldock:
yes! 👍 
side questions: is there a way in core api for this to be detected and corrected? has this already been raised with core services as something we'd like to see fixed?

RasPhilCo (approved):
🏄

msakrejda:
@camillebaldock we used to have the current behavior, and then we changed it to block rollbacks past an add-on attachment to avoid this, but that (believe it or not) caused even more confusion, so we rolled it back.

There's been discussions off and on about separating attachment config vars from user-driven config in order to sidestep these issues entirely but I believe that involves invasive changes so it was always more of a pipe dream project. Happy to sit in if you'd like to revisit this, but I'm not sure who from core services should be involved.

msakrejda:
Also note that similar issues affect anything that requires the config var on any add-on, e.g., `kafka:write`, `kafka:tail`, and `redis:cli`. I believe all the `pg:*` commands all go through this codepath, so this patch should address the issue for both `heroku-pg` and `heroku-pg-extras`, but we should also fix this for Redis and Kafka at some point (though while I see Postgres tickets about this relatively often, I don't believe I've seen tickets for other datastores). Or ideally maybe even unify "get config var value for a specific add-on" into a generic add-on function.

PR #180, opened 2018-01-23
Fix tests and update nock

mehtaphysical:
Some tests where failing because of a bug in nock. Updating nock revealed a bunch of other failing tests.

### DIFF
--- a/test/commands/backups/capture.js
+++ b/test/commands/backups/capture.js
@@ -30,7 +30,7 @@ const shouldCapture = function (cmdRun) {
   it('captures a db', () => {
     addon.app.name = 'myapp'
     api = nock('https://api.heroku.com')
-    api.post('/actions/addon-attachments/resolve', {app: 'myapp', addon_attachment: 'DATABASE_URL'}).reply(200, [{addon}])
+    api.post('/actions/addon-attachments/resolve', {app: 'myapp', addon_attachment: 'DATABASE_URL', addon_service: 'heroku-postgresql'}).reply(200, [{addon}])
 
     pg = nock('https://postgres-api.heroku.com')
     pg.post('/client/v11/databases/1/backups').reply(200, {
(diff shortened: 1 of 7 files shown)


test/commands/backups/capture.js
```diff
@@ -30,7 +30,7 @@ const shouldCapture = function (cmdRun) {
   it('captures a db', () => {
     addon.app.name = 'myapp'
     api = nock('https://api.heroku.com')
-    api.post('/actions/addon-attachments/resolve', {app: 'myapp', addon_attachment: 'DATABASE_URL'}).reply(200, [{addon}])
+    api.post('/actions/addon-attachments/resolve', {app: 'myapp', addon_attachment: 'DATABASE_URL', addon_service: 'heroku-postgresql'}).reply(200, [{addon}])
```
keiko713, line 33: Can you elaborate more on the bug of nock? Also, why we now need to use this `addon_service`?

mehtaphysical: 1) The bug in nock wasn't properly overriding the http request method. So, for example, `test/commands/backups/download.js` was failing:

```
  1) pg:backups:download with no id downloads to latest.dump:
     UnexpectedError:
expected '{"id":"not_found","message":"Not found."}' to equal '{}'

{"id":"not_found","message":"Not found."}
{}

      at cmd.run.then (test/commands/backups/download.js:35:19)
      at <anonymous>
      set UNEXPECTED_FULL_TRACE=true to see the full stack trace

  2) pg:backups:download with id downloads to latest.dump:
     UnexpectedError:
expected '{"id":"not_found","message":"Not found."}' to equal '{}'

{"id":"not_found","message":"Not found."}
{}

      at cmd.run.then (test/commands/backups/download.js:42:19)
      at <anonymous>
      set UNEXPECTED_FULL_TRACE=true to see the full stack trace
```

I think it's this PR in particular that fixes this: https://github.com/node-nock/nock/pull/929.

2) Nock checks to make sure the post body matches the expected body. In the version we are currently using this check isn't enforced (see this issue: https://github.com/node-nock/nock/issues/920). So the `addon_service` is getting posted and we need to add it to our expected post body.

mehtaphysical:
This will be fixed by #173 . I'm going to close this

keiko713:
Oh sad, thanks for the explanation though! Also maybe you wanna delete a branch ;)

PR #181, opened 2018-01-25
Add warning to promote

mehtaphysical:
This adds a warning about potentially failed releases during promotion if release phase is used.

I think there is still an underlying problem with this flow, however. If the `Detach DATABASE` release fails silently and the following `Attach DATABASE` release takes a long time (maybe a long migration) the app will deploy without a database. I think it would be nice to find a way to skip the release for `Detach Database` altogether. Maybe add a flag here to skip release https://github.com/heroku/api/blob/master/lib/api/mediators/attachments/destroyer.rb

### DIFF
--- a/commands/promote.js
+++ b/commands/promote.js
@@ -69,8 +69,13 @@ function * run (context, heroku) {
 
   if (releasePhase) {
     yield cli.action('Checking release phase', co(function * () {
-      let releases = (yield heroku.get(`/apps/${app}/releases`))
-          .sort((a, b) => Date.parse(b.created_at) - Date.parse(a.created_at))
+      let releases = yield heroku.request({
+        path: `/apps/${app}/releases`,
+        partial: true,
+        headers: {
+          'Range': `version ..; max=5, order=desc`
+        }
+      })
       let attach = releases.find((release) => release.description.includes('Attach DATABASE'))
       let detach = releases.find((release) => release.description.includes('Detach DATABASE'))
 
--- a/commands/promote.js
+++ b/commands/promote.js
@@ -7,6 +7,7 @@ function * run (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   const {app, args} = context
   const attachment = yield fetcher.attachment(app, args.database)
+  let hasCurrent = false
 
   yield cli.action(`Ensuring an alternate alias for existing ${cli.color.configVar('DATABASE_URL')}`, co(function * () {
     // Finds or creates a non-DATABASE attachment for the DB currently
@@ -18,6 +19,8 @@ function * run (context, heroku) {
     let attachments = yield heroku.get(`/apps/${app}/addon-attachments`)
     let current = attachments.find(a => a.name === 'DATABASE')
     if (!current) return
+    hasCurrent = true
+
     if (current.addon.name === attachment.addon.name && current.namespace === attachment.namespace) {
       if (attachment.namespace) {
         throw new Error(`${cli.color.attachment(attachment.name)} is already promoted on ${cli.color.app(app)}`)
@@ -49,6 +52,8 @@ function * run (context, heroku) {
   } else {
     promotionMessage = `Promoting ${cli.color.addon(attachment.addon.name)} to ${cli.color.configVar('DATABASE_URL')} on ${cli.color.app(app)}`
   }
+
+  cli.warn('Detaching DATABASE. If you are using the release phase feature (https://devcenter.heroku.com/articles/release-phase), this can cause a failed release.')
 
   yield cli.action(promotionMessage, co(function * () {
     yield heroku.post('/addon-attachments', {
--- a/commands/promote.js
+++ b/commands/promote.js
@@ -53,7 +53,9 @@ function * run (context, heroku) {
     promotionMessage = `Promoting ${cli.color.addon(attachment.addon.name)} to ${cli.color.configVar('DATABASE_URL')} on ${cli.color.app(app)}`
   }
 
-  cli.warn('Detaching DATABASE. If you are using the release phase feature (https://devcenter.heroku.com/articles/release-phase), this can cause a failed release.')
+  if (hasCurrent) {
+    cli.warn('Detaching DATABASE. If you are using the release phase feature (https://devcenter.heroku.com/articles/release-phase), pg:promote can cause a failed release. Please double checkthe releases and make sure the database is properly promoted.')
+  }
 
   yield cli.action(promotionMessage, co(function * () {
     yield heroku.post('/addon-attachments', {
--- a/commands/promote.js
+++ b/commands/promote.js
@@ -75,29 +75,34 @@ function * run (context, heroku) {
       let detach = releases.find((release) => release.description.includes('Detach DATABASE'))
 
       if (!attach || !detach) {
-        throw new Error(`Unable to find Detach and Attach DATABASE releases ${cli.color.app(app)}`)
+        throw new Error('Unable to check release phase. Check your Attach DATABASE release for failures.')
       }
 
+      let endTime = Date.now() + 900000 // 15 minutes from now
       let [attachId, detachId] = [attach.id, detach.id]
       while (true) {
-        let attach = yield heroku.get(`/apps/${app}/releases/${attachId}`)
+        let attach = yield fetcher.release(app, attachId)
         if (attach && attach.status === 'succeeded') {
           let msg = 'pg:promote succeeded.'
-          let detach = yield heroku.get(`/apps/${app}/releases/${detachId}`)
+          let detach = yield fetcher.release(app, detachId)
           if (detach && detach.status === 'failed') {
             msg += ` It is safe to ignore the failed ${detach.description} release.`
           }
           return cli.action.done(msg)
         } else if (attach && attach.status === 'failed') {
           let msg = `pg:promote failed because ${attach.description} release was unsuccessful. Your application is currently running `
-          let detach = yield heroku.get(`/apps/${app}/releases/${detachId}`)
+          let detach = yield fetcher.release(app, detachId)
           if (detach && detach.status === 'succeeded') {
             msg += 'without an attached DATABASE_URL.'
           } else {
-            msg += `with ${current.addon.name} attached as DATABASE_URL`
+            msg += `with ${current.addon.name} attached as DATABASE_URL.`
           }
+          msg += ' Check your release phase logs for failure causes.'
           return cli.action.done(msg)
+        } else if (Date.now() > endTime) {
+          return cli.action.done('timeout. Check your Attach DATABASE release for failures.')
         }
+
         yield new Promise((resolve) => setTimeout(resolve, 5000))
       }
     }))
--- a/commands/promote.js
+++ b/commands/promote.js
@@ -7,7 +7,7 @@ function * run (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   const {app, args} = context
   const attachment = yield fetcher.attachment(app, args.database)
-  let hasCurrent = false
+  let current
 
   yield cli.action(`Ensuring an alternate alias for existing ${cli.color.configVar('DATABASE_URL')}`, co(function * () {
     // Finds or creates a non-DATABASE attachment for the DB currently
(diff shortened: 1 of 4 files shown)


keiko713:
@mehtaphysical you're right about detach/attach thingy. the "hope" here is that people usually run this while they're putting the app in the maintenance mode so there will be no "real" problem other than the app could throw some errors.
The issue should be fixed in the build (release phase) side as Tom mentioned on the card, as well as it would be a nice improvement if the api introduce "detach and attach within one release (like... switch?)".

commands/promote.js
```diff
@@ -50,6 +53,8 @@ function * run (context, heroku) {
     promotionMessage = `Promoting ${cli.color.addon(attachment.addon.name)} to ${cli.color.configVar('DATABASE_URL')} on ${cli.color.app(app)}`
   }
 
+  cli.warn('Detaching DATABASE. If you are using the release phase feature (https://devcenter.heroku.com/articles/release-phase), this can cause a failed release.')
```
keiko713: can it be a bit more generic to `pg:promote`? For example, the real issue here is that the release phase can fail for either detach or attach, or even both. Failing only detach is not that much critical if attach still works. OTOH, failing attach is pretty bad -- it can cause "without a database" situation, or if both detach attach failed, you could keep using the old database without knowing it (as the app would keep "working").
I think it's good to generalize the message and be something like "if you're using the release phase, pg:promote can cause the failure release. please double check the releases and make sure database is properly promoted".

test/commands/promote.js
```diff
@@ -249,3 +240,4 @@ Promoting PURPLE to DATABASE_URL on myapp... done
     return expect(cmd.run({app: 'myapp', args: {}, flags: {}}), 'to be rejected with', err)
   })
 })
+
```
keiko713: extra line

commands/promote.js
```diff
@@ -50,6 +53,10 @@ function * run (context, heroku) {
     promotionMessage = `Promoting ${cli.color.addon(attachment.addon.name)} to ${cli.color.configVar('DATABASE_URL')} on ${cli.color.app(app)}`
   }
 
+  if (hasCurrent) {
+    cli.warn('Detaching DATABASE. If you are using the release phase feature (https://devcenter.heroku.com/articles/release-phase), pg:promote can cause a failed release. Please double checkthe releases and make sure the database is properly promoted.')
```
keiko713: @camillebaldock can we get +1 from product about this? tl;dr, we wanna warn the customer who is using release phase that `pg:promote` can do things that you don't expect. nobody reads warning messages these days but it's better to have than nothing IMO

@mehtaphysical I think we should drop "Detaching DATABASE" part as it can be confusing

msakrejda: Can we avoid showing this to everyone who is not using release phase? It's totally irrelevant to them. If not, what's the release phase uptake? Is it worth warning release phase customers with a message that is not relevant to everyone else?

mehtaphysical: @uhoh-itsmaciek yeah i think that's a good idea. I wasn't sure how to tell if someone was using release phase.

mehtaphysical: maybe we could check if there are any pending or failed releases right after doing the promotion. If there are then we can display the warning?

mehtaphysical: ok so we can either 1) display the warning to all release phase customers or 2) only display to release phase customers if there is a failed release during the detachment phase. I kind of like option 2 better.

keiko713: +1, but also check how trivial it is. I didn't actually expect this to be this wide scope, I'd say 1 is good enough if 2 gets complicated

commands/promote.js
```diff
@@ -61,6 +63,45 @@ function * run (context, heroku) {
       }
     })
   }))
+
+  let releasePhase = (yield heroku.get(`/apps/${app}/formation`))
+    .find((formation) => formation.type === 'release')
+
+  if (releasePhase) {
+    yield cli.action('Checking release phase', co(function * () {
+      let releases = (yield heroku.get(`/apps/${app}/releases`))
```
msakrejda: I'm not sure if this is workable. Shogun (admittedly an extreme case) has 4500+ releases, and paginating through them all (as this will do, 200 at a time) takes over 30s here... I don't know, maybe that's good enough given the alternatives? Or can we assume that the first page of results is almost certain to contain the info we need and stop there? I'm not sure how to opt out of pagination other than [doing everything manually like the api plugin](https://github.com/heroku/heroku-api-plugin/blob/master/src/commands/api.js#L57)--do you folks know off-hand @heroku/cli ?

commands/promote.js
```diff
@@ -61,6 +63,45 @@ function * run (context, heroku) {
...
+          .sort((a, b) => Date.parse(b.created_at) - Date.parse(a.created_at))
+      let attach = releases.find((release) => release.description.includes('Attach DATABASE'))
+      let detach = releases.find((release) => release.description.includes('Detach DATABASE'))
+
+      if (!attach || !detach) {
+        throw new Error(`Unable to find Detach and Attach DATABASE releases ${cli.color.app(app)}`)
```
msakrejda: So in this case, the user would see `Promoting...`, then `Checking release phase`, and then possibly this error? This message doesn't really give context for what's going on here.

commands/promote.js
```diff
@@ -61,6 +63,45 @@ function * run (context, heroku) {
...
+          let msg = `pg:promote failed because ${attach.description} release was unsuccessful. Your application is currently running `
+          let detach = yield heroku.get(`/apps/${app}/releases/${detachId}`)
+          if (detach && detach.status === 'succeeded') {
+            msg += 'without an attached DATABASE_URL.'
+          } else {
+            msg += `with ${current.addon.name} attached as DATABASE_URL`
+          }
+          return cli.action.done(msg)
+        }
+        yield new Promise((resolve) => setTimeout(resolve, 5000))
```
msakrejda: Release phase can (used to be able to?) run for up to 24h, so this could be a while with no feedback. Not that we can really do much.

test/commands/promote.js
```diff
@@ -54,15 +59,13 @@ describe('pg:promote when argument is database', () => {
       confirm: 'myapp'
     }).reply(201)
     return cmd.run({app: 'myapp', args: {}, flags: {}})
-    .then(() => expect(cli.stderr, 'to equal', `Ensuring an alternate alias for existing DATABASE_URL... RED_URL
-Promoting postgres-1 to DATABASE_URL on myapp... done
-`))
+    .then(() => expect(cli.stderr, 'to equal', expectedMessage.replace('PURPLE', 'postgres-1')))
```
msakrejda: I like your thinking here, but honestly I find it easier to work with repeated strings, and copy and paste them from failing test output as necessary (since there it's easy to compare). I don't feel strongly about this, though.

test/commands/promote.js
```diff
@@ -249,3 +239,83 @@ Promoting PURPLE to DATABASE_URL on myapp... done
...
+    return cmd.run({app: 'myapp', args: {}, flags: {}})
+        .then(() => expect(cli.stderr, 'to equal', expectedMessage + 'Checking release phase... pg:promote succeeded.\n'))
+  })
+
+  it('checks release phase for detach failure', () => {
+    api.get('/apps/myapp/releases').reply(200, [{id: 1, description: 'Attach DATABASE'}, {id: 2, description: 'Detach DATABASE'}])
+    api.get('/apps/myapp/releases/1').reply(200, {status: 'succeeded'})
+    api.get('/apps/myapp/releases/2').reply(200, {status: 'failed'})
+    return cmd.run({app: 'myapp', args: {}, flags: {}})
+        .then(() => expect(cli.stderr, 'to equal', expectedMessage + 'Checking release phase... pg:promote succeeded. It is safe to ignore the failed undefined release.\n'))
```
msakrejda: "...safe to ignore the failed undefined release"? 😉

commands/promote.js
```diff
@@ -61,6 +63,45 @@ function * run (context, heroku) {
...
+        let attach = yield heroku.get(`/apps/${app}/releases/${attachId}`)
+        if (attach && attach.status === 'succeeded') {
+          let msg = 'pg:promote succeeded.'
+          let detach = yield heroku.get(`/apps/${app}/releases/${detachId}`)
+          if (detach && detach.status === 'failed') {
+            msg += ` It is safe to ignore the failed ${detach.description} release.`
+          }
+          return cli.action.done(msg)
+        } else if (attach && attach.status === 'failed') {
+          let msg = `pg:promote failed because ${attach.description} release was unsuccessful. Your application is currently running `
```
msakrejda, line 98: So immediately after seeing this, a customer will think, "Ok, how do I fix it?" I know a full explanation is beyond the scope of this PR, but should we give guidance like "Check your release phase logs" or something?

msakrejda (approved):
A few comments, but looks good. I'm not thrilled about adding so much complexity to deal with a totally orthogonal (and... difficult) feature, but I'm glad we found a way to have minimal impact on those not using it. And code itself looks great!

camillebaldock:
hi! just getting round to the initial ping from 6 hours ago; the copy "Detaching DATABASE. If you are using the release phase feature (link to devcenter), this can cause a failed release." is what I had given Ryan this morning, but this PR seems to have taken a different direction. What is the new scope of this and what do you need from product?

msakrejda:
@camillebaldock I think I opened a can of worms with https://github.com/heroku/heroku-pg/pull/181#discussion_r163949892

keiko713:
@uhoh-itsmaciek yes you did 🐛 . I wish Github had more emoji to describe my emotion that I could add to your comment there.
(and I thank you for reviewing this)

commands/promote.js
keiko713: > Or can we assume that the first page of results is almost certain to contain the info we need and stop there?

I believe that the first result definitely should have what we want. I think we can take the same approach of `heroku releases` (not sure where the code is, do you know Maciek?), if I do `heroku releases -n 5`, it only fetches the last 5 releases and that should be pretty enough for us here.

commands/promote.js
keiko713: Yeah this error message can be confusing to the customer. I'd say that we can just ignore this (we didn't have this check before so not showing anything is not degrading), maybe?

commands/promote.js
keiko713: yeah I think we don't wanna that far (explaining everything in this PR), but having "check your release phase logs" sounds like a good thing to add.

commands/promote.js
```diff
@@ -61,6 +63,45 @@ function * run (context, heroku) {
...
+      }
+
+      let [attachId, detachId] = [attach.id, detach.id]
+      while (true) {
+        let attach = yield heroku.get(`/apps/${app}/releases/${attachId}`)
+        if (attach && attach.status === 'succeeded') {
+          let msg = 'pg:promote succeeded.'
+          let detach = yield heroku.get(`/apps/${app}/releases/${detachId}`)
+          if (detach && detach.status === 'failed') {
+            msg += ` It is safe to ignore the failed ${detach.description} release.`
```
keiko713: ❤️

commands/promote.js
mehtaphysical: changed the message to `throw new Error('Unable to check release phase. Check your Attach DATABASE release for failures.')`

commands/promote.js
mehtaphysical: added `' Check your release phase logs for failure causes.'` to the message

commands/promote.js
mehtaphysical: good point. I added a 15 minute timeout.

test/commands/promote.js
mehtaphysical: makes sense. updated

test/commands/promote.js
mehtaphysical: lol ;). fixed

commands/promote.js
msakrejda: Great call @keiko713 -- that code is here: https://github.com/heroku/heroku-apps/blob/master/src/commands/releases/index.js#L22-L30

commands/promote.js
keiko713: uhmm actually does it mean that `pg:promote` command won't finish for up to 15 mins potentially?

msakrejda: Yes #thanksreleasephase

commands/promote.js
mehtaphysical: nice! updated.

jdx: yes, partial: true is what you want

msakrejda:
Looks great, though we should check in with @camillebaldock before shipping.

camillebaldock:
I've had a look, thank you this is very thorough and clear for our customers, you have product thumbs 👍

PR #183, opened 2018-02-02
pg:upgrade --version flag capability added

genslein:
#### What does this PR do?
This is the exposing `--version` flag for enabling upgrades to PostgreSQL 9.6 or 10 from older versions.
#### How has this been tested?

* [x] Specs
* [x] Staging

### DIFF
--- a/commands/upgrade.js
+++ b/commands/upgrade.js
@@ -9,6 +9,7 @@ function * run (context, heroku) {
   const fetcher = require('../lib/fetcher')(heroku)
   let {app, args, flags} = context
   let db = yield fetcher.addon(app, args.database)
+  let target_version = flags.version
 
   if (util.starterPlan(db)) throw new Error('pg:upgrade is only available for follower production databases')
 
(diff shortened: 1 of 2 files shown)


commands/upgrade.js
```diff
@@ -25,8 +25,10 @@ ${cli.color.addon(db.name)} will be upgraded to a newer PostgreSQL version, stop
 
 This cannot be undone.`)
 
+  let body = flags.version ? {host: host(db), body: {version: flags.version}} : {host: host(db), body: {}}
```
msakrejda: This is a little confusing since what you're calling `body` here is really more like "options", with a `body` field. Given that the version is the only thing different, maybe something like

```javascript
let body = {}
if (flags.version) {
  body.version = flags.version
}
yield cli.action(`Starting upgrade of ${cli.color.addon(db.name)}`, co(function * () {
  yield heroku.post(`/client/v11/databases/${db.id}/upgrade`, {host: host(db), body: body})
```
? Not a big deal, though.

msakrejda (approved):

commands/upgrade.js
genslein: I can rename the variable, but we also do something similar to achieve an equivalent result in other commands which is how I modeled it https://github.com/heroku/heroku-pg/blob/master/commands/credentials/create.js#L20 I'm fine with doing either though as long as it's functionally equivalent

PR #184, opened 2018-02-07
Only allow upgrading follower databases

msakrejda:
The Shogun upgrade endpoint has a check for this, but we need to
resolve the URL before we call to Shogun, so check this ourselves.

/cc @mrpaulsmith

### DIFF
--- a/commands/upgrade.js
+++ b/commands/upgrade.js
@@ -18,6 +18,11 @@ function * run (context, heroku) {
   ]
 
   if (status.error) throw new Error(status.error)
+
+  if (!replica.following) {
+    throw new Error('pg:upgrade is only available for follower production databases')
+  }
+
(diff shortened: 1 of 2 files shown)


fnando:
Needs review. cc @fnando @brettgoulder

kvlou (approved):
looks legit

keiko713 (approved):
support folks got confused, would be nice to ship asap 🌮

kwlockwo:
With this change, we should fix our documentation: https://devcenter.heroku.com/articles/upgrading-heroku-postgres-databases#upgrading-with-pg-upgrade

> Unlike with other database upgrade methods, no new instance of your database is created

If it only works on a follower this isn't true.

keiko713:
@kwlockwo can you please leave that comment as a Dev Center feedback? it's a good feedback and would like to see it fixed

PR #185, opened 2018-02-17
add oclif manifest

jdx:


msakrejda (approved):
Seems legit. From the lack of description, I'm assuming there are no user-visible changes?

PR #187, opened 2018-02-22
add warning if follower is behind

mehtaphysical:
add a warning to `pg:unfollow` if the follower is behind


msakrejda:
The warning is not terribly actionable and comes after an irreversible operation... What's our goal here? Can/should we warn before the action? Can we link to more information on follower lag?

PR #191, opened 2018-03-28
Rename getUrl to getConfigVarName

keiko713:
I am a bit crying looking at this code but I've gotta start with somewhere.
This PR is to just simply rename getUrl to getConfigVarName, because the name getUrl is not accurate and super confusing.

I'm hoping to make more PRs on top of this.


msakrejda (approved):
:+1: to starting somewhere

keiko713:
not planning to ship this, but going to merge so that I can make PR on top of this

msakrejda:
@keiko713 yeah, I think that's a great approach. Let's keep going incrementally and then ship a release when we have something interesting.

PR #192, opened 2018-03-29
Add warning when pg:copy-ing from db with additional credentials

camillebaldock:
Part a) of https://trello.com/c/htDenfnh/2182-pgcopy-copyover-does-not-work-with-additional-creds

### DIFF
--- a/commands/copy.js
+++ b/commands/copy.js
@@ -61,6 +61,14 @@ Data from ${cli.color.yellow(source.name)} will then be transferred to ${cli.col
       host: host(attachment.addon)
     })
   }))
+
+  if (source.attachment) {
+    let credentials = yield heroku.get(`/postgres/v0/databases/${source.attachment.addon.name}/credentials`,
+                                     { host: host(source.attachment.addon) })
+    if (credentials.length > 1) {
+      cli.action.warn(`pg:copy will only copy your default credential and the data it has access to. Any additional credentials and data that only they can access will not be copied.`)
+    }
+  }
   yield pgbackups.wait('Copying', copy.uuid, interval, flags.verbose, attachment.addon.app.name)
 }
 
--- a/test/commands/copy.js
+++ b/test/commands/copy.js
@@ -14,9 +14,20 @@ const addon = {
   config_vars: ['DATABASE_URL'],
   plan: {name: 'heroku-postgresql:standard-0'}
 }
+const otherAddon = {
+  id: 2,
+  name: 'postgres-2',
+  app: {name: 'myotherapp'},
+  config_vars: ['DATABASE_URL'],
+  plan: {name: 'heroku-postgresql:standard-0'}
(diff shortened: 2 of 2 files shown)


commands/copy.js
```diff
@@ -51,6 +51,12 @@ Data from ${cli.color.yellow(source.name)} will then be transferred to ${cli.col
   yield cli.action(`Starting copy of ${cli.color.yellow(source.name)} to ${cli.color.yellow(target.name)}`, co(function * () {
     attachment = target.attachment || source.attachment
     if (!attachment) throw new Error('Heroku PostgreSQL database must be source or target')
+    console.log({
+      from_name: source.name,
+      from_url: source.url,
+      to_name: target.name,
+      to_url: target.url
+    })
```
msakrejda: Stray debugging?

msakrejda (changes requested):
Some linting issues, but otherwise looks good.

commands/copy.js
```diff
@@ -61,6 +61,14 @@ Data from ${cli.color.yellow(source.name)} will then be transferred to ${cli.col
       host: host(attachment.addon)
     })
   }))
+
+  if (source.attachment) {
+    let credentials = yield heroku.get(`/postgres/v0/databases/${source.attachment.addon.name}/credentials`,
```
keiko713, line 66: let's only do this when it's shogun resource (aka this won't work for yobuko)

keiko713:
one comment but looks good, did you test in staging?

commands/copy.js
msakrejda: This should work for Yobuko as well--we shipped a similar degenerate credentials API there. But if we know the plan up-front, maybe we should avoid this check, since hobby databases will never have additional credentials? I'm +0 on that.

keiko713: GUESS WHO MADE IT https://github.com/heroku/yobuko/commits/78c42881132e9c9294cc0cb2b20b036463095f7b/web/endpoints/client/postgres.rb 🤦‍♂️  I should have asked Keiko 2017.

Well in that case, yeah I'm +0 as well, it's okay to ship as is.

keiko713 (approved):

camillebaldock:
@keiko713 i tested against two databases doing pg:copy before i added a credential, pg:copy after i added a credential (and also tested what happens when you remove a credential, which it turns out takes a long time, and i found a bug which i am opening a PR in shogun for)

camillebaldock:
@keiko713 going to test on some yobukos now

camillebaldock:
@keiko713 tested with yobuko, behaves as expected (no difference to current behaviour for yobukos)

camillebaldock:
@uhoh-itsmaciek are you happy with these changes now?

msakrejda (approved):
Yes!

PR #194, opened 2018-04-03
Minimize the use of util.getConfigVarName

keiko713:
`util.getConfigVarName` is a bad method that just picks up the index 0 value of config var arrays, of its addon. Addon object has config vars, which tells you what kind of config vars are associated with that addon.
For example, if I have the following addon:

```
$ heroku addons:info postgresql-angular-71290 -a keiko-sushi
=== postgresql-angular-71290
Attachments:  keiko-sushi::HEROKU_POSTGRESQL_CYAN
              keiko-sushi::HEROKU_POSTGRESQL_MAROON
Installed at: Mon Mar 27 2017 07:03:51 GMT-0700 (PDT)
Owning app:   keiko-sushi
Plan:         heroku-postgresql:standard-0
Price:        $50/month
State:        created
```

This has two config vars (attachments), `HEROKU_POSTGRESQL_CYAN_URL` and `HEROKU_POSTGRESQL_MAROON_URL`. In this case, we basically don't know if we get CYAN or MAROON with `util.getConfigVarName`.
Usually, whenever you run `heroku pg` command, you know which attachment you're talking about (e.g. you run `heroku pg:copy DATABASE MAROON`) -- especially when the attachment was given with `fetcher.attachment`. This is important that it respects that (in `pg:copy` case, `MAROON` and not `CYAN`), because sometimes that config vars include creds which makes the situation very confusing if that was mistakenly picked up.

### DIFF
--- a/commands/backups/schedule.js
+++ b/commands/backups/schedule.js
@@ -2,7 +2,6 @@
 
 const co = require('co')
 const cli = require('heroku-cli-util')
-const util = require('../../lib/util')
 
 const TZ = {
   'PST': 'America/Los_Angeles',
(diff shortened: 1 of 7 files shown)


keiko713:
@uhoh-itsmaciek this is something we "kinda paired" long time ago (https://trello.com/c/ZmEiIsFm/729-handle-pg-with-credentials-well-with-pgcopy), let me know if you forget things, happy to explain more!

msakrejda (approved):
Seems legit

PR #195, opened 2018-04-03
Fix double port number display issue

### LINKED ISSUE #193: pg:copy prints the database port twice
When using `pg:copy` I noticed the output printed the database port two times:

```
ᐅ heroku pg:copy postgres://user:pass@server-hostname:5432/db-name RESTORED_DATABASE_URL --app my-app --confirm my-app
Starting copy of database db-name on server-hostname:5432:5432 to RESTORED_DATABASE... done
Copying... done
```

keiko713:
Apparently we were doing something wrong here, `uri.host` is including port number: https://nodejs.org/api/url.html

This will fix https://github.com/heroku/heroku-pg/issues/193

I kinda wanted to ship https://github.com/heroku/heroku-pg/pull/194, but I thought I'd fix one more bug before shipping.

### DIFF
--- a/commands/copy.js
+++ b/commands/copy.js
@@ -14,12 +14,12 @@ function * run (context, heroku) {
   let resolve = co.wrap(function * (db) {
     if (db.match(/^postgres:\/\//)) {
       let uri = url.parse(db)
-      let name = uri.path ? uri.path.slice(1) : ''
-      let hostname = `${uri.host}:${uri.port || 5432}`
+      let dbname = uri.path ? uri.path.slice(1) : ''
+      let host = `${uri.hostname}:${uri.port || 5432}`
       return {
-        name: name ? `database ${name} on ${hostname}` : `database on ${hostname}`,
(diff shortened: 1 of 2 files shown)


msakrejda (approved):

commands/copy.js
```diff
@@ -14,12 +14,12 @@ function * run (context, heroku) {
   let resolve = co.wrap(function * (db) {
     if (db.match(/^postgres:\/\//)) {
       let uri = url.parse(db)
-      let name = uri.path ? uri.path.slice(1) : ''
-      let hostname = `${uri.host}:${uri.port || 5432}`
+      let dbname = uri.path ? uri.path.slice(1) : ''
```
keiko713, line 17: I didn't really bother trying to look into more, but this kinda doesn't make sense that `uri.path` being potentially null  -- we obviously can't `pg:copy` from/to the database that doesn't have dbname with it.
Let's pretend like we didn't see it for now, though.

PR #196, opened 2018-04-17
pass the app name to shogun

coreypurcell:
To allow cross app connection pooling attachments we need the app name.
We were deriving it from the addon's owning app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@heroku-cli/plugin-pg",
   "description": "Heroku CLI plugin to manage Postgres.",
-  "version": "2.9.7",
+  "version": "2.9.6",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@heroku-cli/plugin-pg",
   "description": "Heroku CLI plugin to manage Postgres.",
-  "version": "2.9.6",
+  "version": "2.9.7",
```
msakrejda: This gets bumped separately when we do a release.

msakrejda (changes requested):
The version bump will happen separately on release, but otherwise this looks fine.

coreypurcell:
@uhoh-itsmaciek fixed

msakrejda (approved):

PR #197, opened 2018-04-19
Fix the bug with the attachment name with pg:copy

keiko713:
SOOOO this is confusing. But here is the case:

* there is myapp, with `RED` and `ATTACHED_BLUE`, where `RED`'s billing app is myapp, but `ATTACHED_BLUE` is just an attachment and billing app is myotherapp
* you try to resolve the attachment `ATTACHED_BLUE` with app myapp, it'll success, and will give you something like below format "attachment" object

```
attachment = {
  id: <id of attachment>,
  name: `ATTACHED_BLUE`,
  addon:  {
    id: <id of addon>,
    name: <addon haiku name>,
    app: { id: <id of app>,  name: <billing app name -- so myotherapp> }
  },
  app: { id: <id of app>, name: <attachment attached app name -- so myapp> }
}
```

As you can see, there are two "app" info. When you try to get the config var, it's VERY important to check the config var of "myapp" with `ATTACHED_BLUE`, where we were failing to do so.
This PR fixes it.

### DIFF
--- a/commands/copy.js
+++ b/commands/copy.js
@@ -13,6 +13,7 @@ function * run (context, heroku) {
 
   let resolve = co.wrap(function * (db) {
     if (db.match(/^postgres:\/\//)) {
+      // For the case an input is URL format
       let uri = url.parse(db)
       let dbname = uri.path ? uri.path.slice(1) : ''
       let host = `${uri.hostname}:${uri.port || 5432}`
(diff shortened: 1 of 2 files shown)


keiko713:
(support ticket: https://support.heroku.com/tickets/575823)

msakrejda (approved):
Great find!

PR #198, opened 2018-04-24
Fix pg wait

msakrejda:
Make `pg:wait` wait for a single database correctly (and fix the tests that failed to catch this).

### DIFF
--- a/test/commands/wait.js
+++ b/test/commands/wait.js
@@ -12,8 +12,8 @@ const all = [
 ]
 const fetcher = () => {
   return {
-    all: () => all,
-    addon: () => all[0]
+    all: () => Promise.resolve(all),
+    addon: () => Promise.resolve(all[0])
   }
 }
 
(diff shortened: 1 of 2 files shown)


test/commands/wait.js
```diff
@@ -12,8 +12,8 @@ const all = [
 ]
 const fetcher = () => {
   return {
-    all: () => all,
-    addon: () => all[0]
+    all: () => Promise.resolve(all),
+    addon: () => Promise.resolve(all[0])
```
msakrejda, line 16: These changes ensure that the tests (correctly) fail without the code changes above.

mble (approved):
LGTM 👍

I should learn how this async/await stuff works in Node.

jdx (approved):

jdx:
I think because I moved this from `co` to `async/await` this started to fail. In `co`  you could wait on arrays of promises so `yield [mypromise, myotherpromise]` was the same as `await Promise.all([mypromise, myotherpromise])`

PR #199, opened 2018-04-30
Ensure we wait between pg:wait polling checks

msakrejda:
See https://support.heroku.com/tickets/577414

### DIFF
--- a/commands/wait.js
+++ b/commands/wait.js
@@ -13,7 +13,6 @@ async function run (context, heroku) {
   const notify = require('../lib/notify')(context)
 
   let waitFor = async function waitFor (db) {
-    const wait = require('co-wait')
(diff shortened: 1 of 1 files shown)


commands/wait.js
```diff
@@ -66,7 +65,11 @@ async function run (context, heroku) {
 
       cli.action.status(status.message)
 
-      await wait(interval * 1000)
+      await new Promise((resolve, reject) => {
+        setTimeout(() => {
+          resolve()
+        }, interval * 1000)
+      })
```
msakrejda: No idea if this is the right way to do it, but it looks like `co-wait` is not the right way to do it--it seems to ignore the wait entirely.

idan (approved):
Legit

PR #203, opened 2018-05-10
update deps

jdx:
now we are able to lazy-load commands so it is actually better to
load all of lodash than the smaller modules separately


msakrejda (approved):
:tada:

PR #204, opened 2018-05-14
Add support for -f in pg:psql

### LINKED ISSUE #201: [Feature request] Support -f for the pg:psql command
The `-f` option of `psql` allowing to run a sql file in psql is quite handy when having to write a big query, instead of copy-pasting it in an interactive session. But `heroku pg:psql` does not support it (while it supports `-c`).

My workaround until now was to retrieve the credentials URL and then use `psql` directly. However, this does not work for a private plan DB, as this would not go through the bastion SSH tunnel (and yes, I just went reading the source code of this package to understand how `pg:psql` could connect to the private DB).
It would be great if `-f` could be supported by `heroku pg:psql` directly.

stof:
Closes #201 

The implementation is heavily inspired by the support of `-c`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3689,7 +3689,7 @@ semver-diff@^2.0.0:
   version "5.3.0"
   resolved "https://registry.yarnpkg.com/semver/-/semver-5.3.0.tgz#9b2ce5d3de02d17c6012ad326aa6b4d0cf54f94f"
 
-semver@^5.5.0:
+semver@^5.4.1, semver@^5.5.0:
   version "5.5.0"
   resolved "https://registry.yarnpkg.com/semver/-/semver-5.5.0.tgz#dc4bbc7a6ca9d916dee5d43516f0092b58f7b8ab"
 
--- a/commands/psql.js
+++ b/commands/psql.js
@@ -15,6 +15,8 @@ function * run (context, heroku) {
   cli.console.error(`--> Connecting to ${cli.color.addon(db.attachment.addon.name)}`)
   if (flags.command) {
     process.stdout.write(yield psql.exec(db, flags.command))
+  } else if (flags.file) {
+    process.stdout.write(yield psql.execFile(db, flags.file))
   } else {
     yield psql.interactive(db)
   }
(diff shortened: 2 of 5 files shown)


msakrejda:
Awesome, thank you :heart:! We could probably refactor to reduce some repetition, but this code's pretty stable, so I don't think it's worth it. Anyone else from @heroku/dod-ux care to weigh in? Otherwise I think we can ship this later this week.

msakrejda (approved):

brettgoulder:
@uhoh-itsmaciek I think this code is stable enough, no need for an additional refactor.  We can take a look once we oclif-ify everything.

yarn.lock
```diff
@@ -3689,7 +3689,7 @@ semver-diff@^2.0.0:
   version "5.3.0"
   resolved "https://registry.yarnpkg.com/semver/-/semver-5.3.0.tgz#9b2ce5d3de02d17c6012ad326aa6b4d0cf54f94f"
 
-semver@^5.5.0:
+semver@^5.4.1, semver@^5.5.0:
```
msakrejda, line 3692: Do you know why this yarn.lock change is here @stof? Can you drop it? Even if necessary, this does not seem relevant to this patch.

msakrejda (changes requested):

yarn.lock
stof: that was done by `yarn`. Maybe I'm using a newer version of Yarn than when this lock file was generated, with a bug fixed in it.

I inspected the lock file, and there is indeed a package depending on `semver@^5.4.1`. So this change totally make sense. I suspect a bug in an older version of Yarn making it disappear.

stof: I can of course drop it, but any contributor running `yarn` locally to install the dependencies will re-submit it again.

msakrejda: Let me see if I can upgrade my yarn locally and repro the change separately.

msakrejda: Yep, you're right--in fact this is already in `master` as part of another patch that landed earlier. I just wanted to make sure to keep lockfile changes separate, but it's moot. Merging shortly!

msakrejda (approved):
Thanks!

msakrejda:
@stof I published a new plugin version that includes this and we should have it in a CLI update in the next few business days. Thanks for the patch!

PR #206, opened 2018-05-18
Adding try/catch to attachment filtering

binarycleric:
This fixes a bug that causes `pg:psql` and other commands to break if an addon is created that hasn't yet exported config variables. This is a rather uncommon situation but a provisioning bug shouldn't prevent us from using most `pg:*` commands.

### DIFF
--- a/lib/fetcher.js
+++ b/lib/fetcher.js
@@ -58,7 +58,18 @@ module.exports = heroku => {
         throw new Error(`${cli.color.app(app)} has no databases`)
       }
 
-      matches = attachments.filter(attachment => config[db] && config[db] === config[pgUtil.getConfigVarName(attachment.config_vars)])
+      matches = attachments.filter(attachment => {
+        let configVarName
+
+        try {
+          configVarName = pgUtil.getConfigVarName(attachment.config_vars)
+        } catch (error) {
+          debug(`found an attachment without any exported config_vars: ${attachment.name}`)
+          configVarName = null
+        }
+
+        return config[db] && config[db] === config[configVarName]
+      })
 
       if (matches.length === 0) {
         let validOptions = attachments.map(attachment => pgUtil.getConfigVarName(attachment.config_vars))
--- a/test/lib/fetcher.js
+++ b/test/lib/fetcher.js
@@ -282,6 +282,47 @@ describe('fetcher', () => {
         }))
       })
 
+      it('returns DATABASE_URL if another addon has a bad attachment', () => {
+        const err = new Error()
+        err.statusCode = 404
+        err.body = {id: 'not_found'}
+        err.message = 'Not Found'
+
+        stub.withArgs(sinon.match.any, 'myapp', 'attach-app::DATABASE_URL').returns(Promise.reject(err))
+
+        api.get('/apps/attach-app/config-vars').reply(200, {
+          'DATABASE_URL': 'postgres://pguser:pgpass@pghost.com/pgdb',
+          'HEROKU_POSTGRESQL_PINK_URL': 'postgres://pguser:pgpass@pghost.com/pgdb'
+        })
+
+        let plan = {name: 'heroku-postgresql:hobby-dev'}
+        let attachments = [
+          {
+            app: {name: 'attach-app'},
+            addon: {id: 100, name: 'postgres-1', plan},
+            config_vars: ['HEROKU_POSTGRESQL_PINK_URL']
+          },
+          {
+            app: {name: 'attach-app'},
+            addon: {id: 100, name: 'postgres-1', plan},
+            config_vars: []
+          }
+        ]
+
+        api.get('/apps/attach-app/addon-attachments').reply(200, attachments)
+
+        return fetcher(new Heroku()).database('myapp', 'attach-app::DATABASE_URL')
+        .then((db) => expect(db, 'to equal', {
+          user: 'pguser',
+          password: 'pgpass',
+          database: 'pgdb',
+          host: 'pghost.com',
+          port: null,
+          attachment: attachments[0],
+          url: url.parse('postgres://pguser:pgpass@pghost.com/pgdb')
+        }))
+      })
+
       it('returns when DATABASE db arg', () => {
         const err = new Error()
         err.statusCode = 404


msakrejda:
@binarycleric what are the repro steps for this? I just tried to reproduce and I get the error

```
 ▸    No config vars found for DATABASE; perhaps they were removed as a side effect of heroku rollback? Use heroku addons:attach to create a new attachment and then heroku addons:detach to remove the current
 ▸    attachment.
```

which does not exactly make it clear what the issue is, but it's different than the error I recall you were seeing.

jdx:
I have no idea what the right thing to do here is. @heroku/ecosystem, would you mind lending some thoughts?

msakrejda:
@heroku/dod-ux this may be of interest as well: right now for commands where config vars are required (like `pg:psql`), we need to abort when they are not present. The current warning highlights one likely source of this (`heroku rollback`), but there's at least one other situation that can cause it: the add-on is still provisioning (can take a while especially for large forks or followers) and the config vars are not yet present. There may be others but that's all I can think of for now. Ideally, we'd have a consistent mechanism for handling this across add-ons, but if not, we should at least be consistent within data add-ons.

mathias:
https://github.com/heroku/heroku-cli-addons/blob/master/lib/addons_wait.js#L12 -- `addon.state === 'provisioning'` ?

msakrejda:
@mathias good call; we probably don't want to wait in this case, but we can def check the state. I think there's still an edge case because the provisioning time limit means Shogun sometimes tells users the add-on is ready even when it's not, but I think we can mostly ignore that; it should be pretty rare... 

There's also a potential permissions issue: you may have access to an attachment without having access to the billing app, so you may be able to view the auxiliary attachment config vars without being able to hit `/addons/<the-addon>` to view its provisioning state, but maybe not worth worrying about.

mathias:
Yes, I didn't expect we needed to wait, I just wanted to point out how to detect it. That is only if it is using the asysnc provisioning workflow, though.

We don't have cross-owner attachments in most cases, so I am not sure that the attachment to a billing app you can't see is that big an issue.

mikehale:
Could they check for `status == 'provisioned'` instead? That should work for both sync, and async add-ons, as they both have that status when they are ready.

mathias:
Yes, that's even better. Good catch @mikehale!

msakrejda:
>We don't have cross-owner attachments in most cases, so I am not sure that the attachment to a billing app you can't see is that big an issue.

Well, it does not necessarily needs to be cross-owner for this to come into play, right? A team can own both apps but users within the team may have different levels of access, no? This is the only way to control access to HPG Credentials using the platform permissions model...

mathias:
Being a member of the team should give them `view` on both apps. That should be enough to see both sides.

msakrejda:
@mathias you're right--I think I was conflating actual access to config vars (for psql) with access to add-on metadata. So there may still be edge cases with one-off configurations, but the common case should not be an issue.

binarycleric:
> Could they check for status == 'provisioned' instead?

fwiw, it doesn't look like `attachment.addon` has state in this case.  Here's an example of what's in `attachment`.

```
{ addon: 
   { id: 'd978d256-3016-4503-9d94-9726dc6b3556',
     name: 'postgresql-jdaniel-rigid-57849',
     app: 
      { id: '4264af07-6034-43a7-b8de-50bd3e4860ea',
        name: 'shogun-jdaniel' },
     plan: 
      { id: '79b60b8a-3366-474e-b19e-b6ae98dbf545',
        name: 'heroku-postgresql-jdaniel:private-2' } },
  app: 
   { id: '4264af07-6034-43a7-b8de-50bd3e4860ea',
     name: 'shogun-jdaniel' },
  id: 'cc56792d-2c91-45eb-9153-960a10e4e73f',
  name: 'DATABASE',
  namespace: null,
  created_at: '2018-04-10T19:31:57Z',
  updated_at: '2018-04-10T19:31:57Z',
  web_url: 'https://addons-sso.heroku.com/apps/4264af07-6034-43a7-b8de-50bd3e4860ea/addons/d978d256-3016-4503-9d94-9726dc6b3556',
  config_vars: 
   [ 'DATABASE_URL',
     'DATABASE_BASTIONS',
     'DATABASE_BASTION_KEY',
     'DATABASE_BASTION_REKEYS_AFTER' ] }
```

mathias:
Are you hitting some internal endpoint to get this? The [Platform API response](https://devcenter.heroku.com/articles/platform-api-reference#add-on) has it.

mathias:
Oh. You request an attachment and somehow expand an add-on record in it? That may not include all the attrs. We may have to adjust that.

binarycleric:
It is using `appAttachment` from [heroku-cli-addons](https://github.com/heroku/heroku-cli-addons) https://github.com/heroku/heroku-cli-addons/blob/master/lib/resolve.js#L176

Do you mean adjust it client-side or server-side? I'd worry about making n+1 API requests here since it'll be painfully slow for customers with a large number of addons.

msakrejda:
So I think further `provisioned` state checks are worthwhile, but we're getting ahead of ourselves--sorry I derailed the discussion. My main issue here is that I was not able to reproduce the original error (I got a different error, more accurate, error which I don't think warrants these changes). Do you have a reproducible test case (or reproduction steps) @binarycleric?

mathias:
@binarycleric It looks like it is using an Inclusion header here:
https://github.com/heroku/heroku-pg/blob/b6b78bdfe0cf33fcb5bc40b99b95519aa4b3612d/lib/fetcher.js#L111

Those parts of the API don't necessarily follow the Platform API response schema -- the embedded records are simplified for smaller responses. I think we forgot to add `state` there. So this would be a change in Core API.

binarycleric:
Not at anymore @uhoh-itsmaciek. This happened because of an error in my staging environment with addon provisioning and I provisioned the broken addon once I realized what had happened. I had a postgres follower that didn't quite make it and left `attachment.config_vars` empty which triggered the error in `getConfigVarName`.

I could try and reproduce it in my staging environment with some hacks but I feel like the issue still stands. A broken addon shouldn't be able to block access to `pg:*` commands across a whole app.

msakrejda:
In that case should we make `getConfigVarName` not throw instead (and, e.g., return `null`)?

binarycleric:
@uhoh-itsmaciek I was thinking something like that too but I didn't want to make that change at first since I'm not too familiar with how all this works. I can give it a shot and see if anything breaks. I also found the serializer in Core API and adding `state` doesn't look like it would be too difficult.

mathias:
The only thing that comes to mind is, any other commands that depend on `config_vars` getting set will also need to handle null/`[]` case.

binarycleric:
Agreed @mathias. I'd rather err on the side of caution for now. Having `state` in Core API is probably better. I'm working on a branch that'll be up soon.

binarycleric:
@uhoh-itsmaciek Alright, I was able to reproduce in my staging environment. Basically, I just provisioned a follower against a test database and I'm not locked out of `heroku pg:psql` and other commands.

Here's the attachment object in question.

```
{ addon: 
   { id: 'e25e6fb4-f778-4491-9e96-23a9540446ef',
     name: 'postgresql-jdaniel-rigid-67883',
     app: 
      { id: '4264af07-6034-43a7-b8de-50bd3e4860ea',
        name: 'shogun-jdaniel' },
     plan: 
      { id: '6ec66f22-01fb-4871-8d8e-7dc156872564',
        name: 'heroku-postgresql-jdaniel:standard-0' } },
  app: 
   { id: '4264af07-6034-43a7-b8de-50bd3e4860ea',
     name: 'shogun-jdaniel' },
  id: '66240257-b292-42c1-be87-8fb7121a29d5',
  name: 'HEROKU_POSTGRESQL_JDANIEL_CYAN',
  namespace: null,
  created_at: '2018-05-21T15:59:11Z',
  updated_at: '2018-05-21T15:59:11Z',
  web_url: 'https://addons-sso.heroku.com/apps/4264af07-6034-43a7-b8de-50bd3e4860ea/addons/e25e6fb4-f778-4491-9e96-23a9540446ef',
  config_vars: [] }
```

binarycleric:
🤔 I worry about giving `heroku-pg` too much information about an addon. I feel like in cases of `heroku pg:psql` it should only know if the config var is present or not. Or maybe I am thinking about this the wrong way.

mathias:
Sounds good. Adding the `addon.state` everywhere an add-on is serialized into another record would require a lot of schema changes (but it is additions, so not a big deal) and the association is already loaded there, so it isn't looking up anything it didn't already have.

tl;dr adding `addon.state` is something we can do, and it isn't terribly hard or a performance concern, but if we don't have to change Platform API, that's even easier.

msakrejda:
Let's take a step back:

 1. What is the original error?
 2. What would we ideally like to happen in this situation?
 3. Is there a less ideal solution that is considerably less work (especially important if (1) is not common)?

So (1): the code [here](https://github.com/heroku/heroku-pg/blob/c0fc9a95051ef43e5b34fc55abd73542da35b1da/lib/util.js#L12) throws because it can't find a config var because there are no config vars at all at this point, right?

For (2): ...tell the user that the add-on has not yet finished provisioning?

I'm going to refrain from commenting on (3) until I make sure I understand the first two correctly.

binarycleric:
@uhoh-itsmaciek Sorry if this went off the rails. Maybe I haven't done a great job at explaining the issue.

Here is the original error from after a provisioned a follower that has yet to export config vars. I've only reproduced this in staging, which has less workers and less-powerful dynos, so it may be a non-issue in production.

Provisioned the follower

```bash
heroku addons:create heroku-postgresql-jdaniel:standard-0 --version 9.5 --follow leader-name-here
```

Then went to `pg:psql` into the app's main database, which is the default with `DATABASE_URL`. This failed.

```bash
➜  shogun git:(pg_upgrade_increase_timeout) h pg:psql
 ▸    Database URL not found for this addon
```

After the follower finishes provisioning, `pg:psql` works again for `DATABASE_URL`.

```
heroku-postgresql-jdaniel (postgresql-jdaniel-rigid-67883)     standard-0  free        created
 └─ as HEROKU_POSTGRESQL_JDANIEL_CYAN
```

```bash
➜  shogun git:(pg_upgrade_increase_timeout) h pg:psql
--> Connecting to postgresql-jdaniel-rigid-57849
psql (10.1, server 10.2 (Ubuntu 10.2-1.pgdg16.04+1))
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

shogun-jdaniel::DATABASE=>
```

---

Ideally, commands like `pg:psql` shouldn't break for all addons if a single one hasn't exported config vars yet. After provisioning the follower, I can't `pg:psql` into `DATABASE_URL` until the follower has exported config vars.

binarycleric:
Wait... This is likely me not understanding the code very well but I think I realized the problem. Explanation incoming.

binarycleric:
Okay, this looks like an issue with staging. `matchesHelper` keeps falling through since the database isn't being found by `/actions/addon-attachments/resolve`.

🤦‍♀️

binarycleric:
Closing since I realized this is only a thing that can happen in staging.

PR #207, opened 2018-05-18
fix: force exit after psql closes

jdx:


commands/psql.js
```diff
@@ -17,6 +17,9 @@ function * run (context, heroku) {
     process.stdout.write(yield psql.exec(db, flags.command))
   } else {
     yield psql.interactive(db)
+    // sometimes psql hangs and causes the stdout "drain" event not to be fired in cli-ux
+    // so we force an exit here
+    cli.exit(0)
```
msakrejda, line 22: So the issue is that the CLI is waiting for a stdout event that is not coming? That is, we want to ensure that once psql exits, the command is done and any pending background processing is moot?

msakrejda (approved):

commands/psql.js
jdx: yeah exactly. It's strange that this doesn't have anything to do with stdout really, the issue is that somewhere in the code path some node task is starting and not quitting

PR #208, opened 2018-05-26
add terminate flag to kill

mehtaphysical:
Right now with `heroku sudo` the force flag is not sent because sudo also uses force. This PR adds a terminate flag as a workaround.

### DIFF
--- a/commands/kill.js
+++ b/commands/kill.js
@@ -9,12 +9,12 @@ function * run (context, heroku) {
 
   const {app, args, flags} = context
   const {pid, database} = args
-  const {force} = flags
+  const {force, terminate} = flags
 
   let db = yield fetcher(heroku).database(app, database)
 
   let query = `
-SELECT ${force ? 'pg_terminate_backend' : 'pg_cancel_backend'}(${parseInt(pid)});`
+SELECT ${force || terminate ? 'pg_terminate_backend' : 'pg_cancel_backend'}(${parseInt(pid)});`
 
   let output = yield psql.exec(db, query)
   process.stdout.write(output)
@@ -26,7 +26,7 @@ module.exports = {
   description: 'kill a query',
   needsApp: true,
   needsAuth: true,
-  flags: [{name: 'force', char: 'f'}],
+  flags: [{name: 'force', char: 'f'}, {name: 'terminate', char: 't'}],
   args: [
     {name: 'pid'},
     {name: 'database', optional: true}


mehtaphysical:
[heroku sudo command doesn't pass flags](https://trello.com/c/6TUssHmx/839-heroku-sudo-command-doesnt-pass-flags)

msakrejda (changes requested):
This seems to be more of heroku sudo problem than a heroku-pg problem. I would really like to avoid changing customer-facing commands to work around the limitations of an internal plugin.

If there is no other way to work around this, can we hide the second flag so customers don't see two different flags for the same thing?

mehtaphysical:
@uhoh-itsmaciek for sure. Which way would you prefer to do it? Also, @tcrayford

msakrejda:
Well, it looks like a bunch of other commands have `--force` flags:

```
msakrejda@localhost:~$ for command in $(h commands); do h help $command | sed "s/^/$command: /"; done | grep force
addons:destroy:   -f, --force            allow destruction even if connected to other apps
addons:remove:   -f, --force            allow destruction even if connected to other apps
builds:push:   -f, --force          force push to overwrite remote branch
logs:   --force-colors       force use of colors (even on non-tty output)
pg:credentials:rotate:   --force                forces rotating the targeted credentials
pg:kill:   -f, --force
pg:maintenance:run:   -f, --force
rake:   --no-tty             force the command to not run in a tty
redis:maintenance:   -f, --force          start maintenance without entering application
run:   --no-tty             force the command to not run in a tty
sudo:   -f, --force
```

So other than the bogus matches of `logs`, `run`, `rake`, and `sudo`, you'd need to change all these other commands to work around the issue (and that's just for the plugins that I have installed). Can we change `sudo` instead? Any generic flags on that command are going to have a bunch of awkward conflicts.

Or does oclif support `--`? We could do something like `heroku sudo --force -- pg:kill --force` then...

msakrejda:
Any thoughts on the comment above @jdxcode?

jdx:
I agree with @uhoh-itsmaciek and this is why we have the `--` argument. Just use `heroku sudo -u someuser -- pg:kill -f`.

If you read the help for `heroku help sudo` it always uses the `--` for this reason.

msakrejda:
Oh I had no idea this works already. Neat. In that case, I think we can consider this moot? I'm closing; please reopen if you hit issues with `--`.

PR #209, opened 2018-05-30
autodoc on releases

jdx:


commands/credentials/destroy.js
```diff
@@ -41,10 +41,7 @@ module.exports = {
   description: 'destroy credential within database',
   needsApp: true,
   needsAuth: true,
-  help: `Example:
-
-    heroku pg:credentials:destroy postgresql-transparent-56874 --name cred-name -a woodstock-production
-`,
+  examples: '$ heroku pg:credentials:destroy postgresql-transparent-56874 --name cred-name -a woodstock-production',
```
msakrejda, line 44: @jdxcode i noticed that some of our examples include `-a ...`, and some assume it's implicit. Is there a convention for CLI examples re: this?

msakrejda (approved):

commands/credentials/destroy.js
jdx: there isn't and we should settle on a standard. My initial thought is to leave it out as the app name is already specified in the help

jdx: I feel `-a` is so common in the CLI the users learn it pretty quickly

msakrejda: Yeah, makes sense to me. You can leave as is for now; I'll try to drop `-a` in the future.

